PR #30, opened 2020-08-31
CSVLoader errors on double newlines at the end of files

Craigacp:
Which would be fine (as it's not really to spec) if it wasn't for the fact that some datasets (like the Irises we use in the tutorial) come with a double new line at the end.

This patch treats a double newline as the end of the file. Previously if it was in the middle of the file it would cause it to error out, and if it was the end of the file it would also error out.


Craigacp:
Now it logs a warning for extraneous blank lines and keeps reading the file.

PR #34, opened 2020-09-02
Samanthacampo/kmeans++

samanthacampo:
Add kmeans++ centroid initialization option to kmeans.

### DIFF
--- a/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
+++ b/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
@@ -34,10 +34,15 @@ import org.tribuo.provenance.TrainerProvenance;
 import org.tribuo.provenance.impl.TrainerProvenanceImpl;
 import org.tribuo.util.Util;
 
-import javax.xml.stream.XMLInputFactory;
 import java.time.OffsetDateTime;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.SplittableRandom;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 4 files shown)


Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
```diff
@@ -284,9 +310,117 @@ public int getInvocationCount() {
...
+     * @param numFeatures The number of features.
+     * @return A {@link DenseVector} containing the information from vec.
+     */
+    private static DenseVector sparseToDense(SparseVector vec, int numFeatures) {
+        DenseVector dense = new DenseVector(numFeatures);
+        dense.intersectAndAddInPlace(vec);
+        return dense;
+    }
+
+    private static double getDistance(DenseVector cluster, SGDVector vector,
```
Craigacp, line 402: Bit of javadoc for this please.

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/TrainTest.java
```diff
@@ -54,6 +55,9 @@ public String getOptionsDescription() {
         public int iterations = 10;
         @Option(charName='d',longName="distance",usage="Distance function to use in the e step. Defaults to EUCLIDEAN.")
         public Distance distance = Distance.EUCLIDEAN;
+        @Option(charName='i',longName="initialisation",usage="Type of initialisation to use for centroids. Defaults " +
```
Craigacp, line 58: There already is a `i` argument, so this options class is invalid, could you change it to something else?

Craigacp: Also, there's no need to put the default into the usage string, I updated the options code to print that automatically. While you're in here could you remove the default bit from the other usage strings? I missed that in the docs pass I did.

Craigacp (changes requested):
Few small changes, mostly documentation.

Craigacp (approved):
LGTM

PR #35, opened 2020-09-04
Disable timestamps on javadoc build so that diff isn't confused

jhalexand:
Building the checked-in javadoc for the web site causes changes in every file since javadoc includes the timestamp. Disabling the timestamp inside each html file prevents this.


Craigacp (approved):
LGTM

Issue #58, opened 2020-10-11
MNIST is not loading in Config.json


nikkisingh111333:
i m using tribuo and trying to find mnist dataset in config json ...all the components are loading nut MNIST test/train is giving me NULL POINTER EXCEPTION. 
1.my config code
` ConfigurationManager.addFileFormatFactory(new JsonConfigFactory());
   String configFile = "config.json"; 
   ConfigurationManager cm = new ConfigurationManager(configFile);
   System.out.println( cm.lookup("mnist-train"));
    //GIVING NULL POINTER EXCEPTION
     System.out.println( cm.lookup("logistic")); //RUNNING PERFECTLY FINE RETURNING LOGISSTIC TRAINER `
2.MY CONFIG.JSON 
![Screenshot (37)](https://user-images.githubusercontent.com/49682940/95676540-63341380-0bdc-11eb-87a3-309df0c71c3a.png)
 whats the matter?

neomatrix369:
I got a similar one in the Configuration Tutorial, see attached HTML file, has multiple errors including one for MNist:
```
com.oracle.labs.mlrg.olcut.config.PropertyException: Component: mnist-train, RuntimeException thrown by postConfig (caused by null)
	at com.oracle.labs.mlrg.olcut.config.PropertySheet.getOwner(PropertySheet.java:332)
	at com.oracle.labs.mlrg.olcut.config.PropertySheet.getOwner(PropertySheet.java:239)
	at com.oracle.labs.mlrg.olcut.config.ConfigurationManager.innerLookup(ConfigurationManager.java:1374)
	at com.oracle.labs.mlrg.olcut.config.ConfigurationManager.lookup(ConfigurationManager.java:1357)
	at com.oracle.labs.mlrg.olcut.config.ConfigurationManager.lookup(ConfigurationManager.java:1300)
	at .(#37:1)
```
 
[configuration-tribuo-v4-notebook-errors.html.txt](https://github.com/oracle/tribuo/files/5361125/configuration-tribuo-v4-notebook-errors.html.txt)
_(drop .txt in the name of attached file)_

Craigacp:
No I think that's the same one. I get this stack trace out if the file it's trying to load don't exist:
```
Property Exception component:'train' - RuntimeException thrown by postConfig (caused by null)
com.oracle.labs.mlrg.olcut.config.PropertyException: Component: train, RuntimeException thrown by postConfig (caused by null)

[...]

PR #61, opened 2020-10-11
Makes IDXDataSource throw FileNotFoundException when the files don't exist

Craigacp:
### Description
If the idx files passed into `IDXDataSource` don't exist then it throws a random looking `NullPointerException` out of readData. This obscures what's actually going wrong, and it really should throw `FileNotFoundException`, so now it does.

### Motivation
Fixes the error message in #58 .


JackSullivan (approved):
Looks good to me

PR #66, opened 2020-10-19
Adding an OffsetDateTimeExtractor, deprecating a DateExtractor constructor

Craigacp:
### Description
Adds a `FieldExtractor` that can process date times to go with the one that can process dates. This one uses `OffsetDateTime` to capture timezone information. It's possible we'll need a `LocalDateTime` one for fields without time zone information as `OffsetDateTime` requires it and will fail to parse otherwise.

Also it deprecates the `DateExtractor` constructor which accepts a `DateTimeFormatter` as it's not possible to recover the format string out of such an object, and so the provenance isn't convertible into a configuration. It does however add the toString of the formatter which should be useful for figuring out what the original date format string was even if it can't be automatically generated.

### Motivation
Fixes a bug in the `DateExtractor`s provenance and adds a timestamp capability to the `RowProcessor`s metadata extraction.

### DIFF
--- a/Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
@@ -50,20 +50,20 @@ public class DateExtractor extends SimpleFieldExtractor<LocalDate> {
         postConfig();
     }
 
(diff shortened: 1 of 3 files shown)


JackSullivan (approved):
Looks good to me

PR #69, opened 2020-10-20
Ignore a UTF-8 byte order mark in the csv package

### LINKED ISSUE #65: CSVDataSource ignoring first column
The first column in my CSV file is always ignored i.e. doesn't get loaded into the feature map.

Example: if Role is the first column in the file, it doesn't appear when the feature map is logged. If I insert a "dummy" column, then role appears.

```
        Map<String, FieldProcessor> fieldProcessorMap = new HashMap<>();

        fieldProcessorMap.put("YearsExperience", new DoubleFieldProcessor("YearsExperience") );
        fieldProcessorMap.put("Location", new IdentityProcessor("Location") );
        fieldProcessorMap.put("Role", new IdentityProcessor("Role"));

        var responseProcessor = new FieldResponseProcessor<>("CurrentSalary", "0", new LabelFactory());

        var id = new IntExtractor("ID");

        var rowProcessor = new RowProcessor<>(Collections.singletonList(id), null, responseProcessor, fieldProcessorMap, Collections.emptySet());

        var csvDataSource = new CSVDataSource<>(Paths.get("data/dummy/sample1.csv"), rowProcessor, true);

        var dataSplitter = new TrainTestSplitter<>(csvDataSource, 0.7, 1L);

        var trainingDataset = new MutableDataset<>(dataSplitter .getTrain());
        var testingDataset = new MutableDataset<>(dataSplitter .getTest());

        for (var i : trainingDataset.getFeatureMap())
            log.info(i);
```

Craigacp:
### Description
This change reads the first few bytes into a `PushbackReader`, checks if they are a byte order mark for UTF-8 and discards them if they are. If they are regular bytes they are pushed back onto the reader and loading proceeds as normal.

The UTF-8 byte order mark is unnecessary as UTF-8 doesn't have ordered bytes. Tribuo doesn't support non-UTF-8 file formats in it's data loading.

### Motivation
If there is a byte order mark in a UTF-8 CSV file (as frequently inserted by Microsoft Excel), then OpenCSV will stitch it to the front of the first column, causing the name based parsing in `RowProcessor` among other places to fail. OpenCSV doesn't ignore these by default (unlike other CSV parsers) so here we patch the reader to do so.

Fixes #65.

### DIFF
--- a/Data/src/main/java/org/tribuo/data/csv/CSVDataSource.java
+++ b/Data/src/main/java/org/tribuo/data/csv/CSVDataSource.java
@@ -35,6 +35,10 @@ import org.tribuo.data.columnar.RowProcessor;
 import org.tribuo.provenance.ConfiguredDataSourceProvenance;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.PushbackInputStream;
+import java.io.PushbackReader;
+import java.io.Reader;
 import java.net.URI;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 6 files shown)


JackSullivan (approved):
Looks good to me

PR #76, opened 2020-10-26
Fixing a bug in ONNX LabelTransformer

Craigacp:
### Description
The check was the wrong way around when converting output probability matrices from pytorch models. This change fixes it and adds a unit test based on a simple LeNet style CNN.


jhalexand (approved):
That test method name is truly amazing ;-)

Issue #77, opened 2020-10-28
CSVLoader.loadDataSource() Response not found


Eyal-Elbaz:
CSVLoader.loadDataSource() Response not found

`String[] feature_names = {"LotArea", "YearBuilt", "1stFlrSF", "2ndFlrSF","FullBath", "BedroomAbvGr", "TotRmsAbvGrd"};
var regressionFactory = new RegressionFactory();
var csvLoader = new CSVLoader<>(',', regressionFactory);
var houseDataSource = csvLoader.loadDataSource(Paths.get("train.csv"), "SalePrice", feature_names);`

this code results:

> Exception in thread "main" java.lang.IllegalArgumentException: Response SalePrice not found in file train.csv

[link to dataframe](https://www.kaggle.com/c/house-prices-advanced-regression-techniques/data?select=train.csv)

Craigacp:
If you're using the `CSVLoader` you have to use the full set of columns in the file and it will read all of them. Your `feature_names` variable is treated by the `CSVLoader` as the headers of the csv file, which are already present in that csv so you don't need to supply it. The error message is because the supplied headers don't include the column you chose as the response. 

If as I think you're intending to do, you want to use a subset of the columns in that csv as features then you need to use `CSVDataSource`. You can look at the new columnar data tutorial which explains how to use the columnar processing infrastructure to load anything which isn't a csv of floats. It'll be merged into main shortly - https://github.com/oracle/tribuo/blob/doc-fixes/tutorials/columnar-tribuo-v4.ipynb, at that point you'll be able to find it in the tutorials folder and on the website when the 4.0.2 release happens.

Issue #60, opened 2020-10-11
[Docs] Discrepancy in Gradle instruction between docs on Tribuo.org and GitHub README.md


neomatrix369:
**Describe the bug**

The Gradle tab in https://tribuo.org/#gradle says:

```gradle
implementation ("org.tribuo:tribuo-all:4.0.1@pom") {
    transitive = true // for Groovy
    // isTransitive = true // for Kotlin
}
```

while the README on GitHub says:

```
api 'org.tribuo:tribuo-all:4.0.0@pom'
```

The former works in Gradle 6.5, the latter does not work, and gives the below error:

```
* What went wrong:
A problem occurred evaluating root project 'tribuo-classification'.
> Could not find method api() for arguments [org.tribuo:tribuo-all:4.0.1@pom] on object of 
type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

```

**Expected behaviour**

Both docs to refer to the same information. Also would help if there was a bit more context to the Gradle docs, i.e.:
```gradle
dependencies {
    implementation ("org.tribuo:tribuo-all:4.0.1@pom")
    ....
}
```

Craigacp:
The latter should work if you've got `java-library` imported (aside from the fact that Gradle doesn't transitively pull in pom dependencies). Gradle's split between `api` and `implementation` is frustrating as it makes it difficult to give exact snippets because they aren't defined in the same plugin. I've updated the readme in the [docs-fixes](https://github.com/oracle/tribuo/tree/doc-fixes) branch where we're aggregating these small fixes.

neomatrix369:
> The latter should work if you've got `java-library` imported (aside from the fact that Gradle doesn't transitively pull in pom dependencies). Gradle's split between `api` and `implementation` is frustrating as it makes it difficult to give exact snippets because they aren't defined in the same plugin. I've updated the readme in the [docs-fixes](https://github.com/oracle/tribuo/tree/doc-fixes) branch where we're aggregating these small fixes.

That's fine, for now, I'm only trying to get things to work and reporting issues whether I have a workaround or not.

Thnx for the quick response. I will try to improve on this later on.

neomatrix369:
I have an example Gradle settings file we could link from the Docs, see https://github.com/neomatrix369/awesome-ai-ml-dl/blob/master/examples/tribuo/build.gradle

Or you could craft a pseudo/example version from this one. It an extra step of creating uberjars but that's something we can strip off if you like

[...]

Issue #53, opened 2020-10-02
CSVLoader.loadDataSource() cannot load data without objective variable


k-tamura:
**problem**

`CSVLoader.loadDataSource()` cannot load  data which does not have objective (response) variable. For example:

```java
// Load Titanic test data downloaded from Kaggle (https://www.kaggle.com/c/titanic)
ListDataSource dataource = csvLoader.loadDataSource(Paths.get("test.csv"),"Survived");
List<Prediction> predicts = model.predict(dataource);
```

this code results:

```
Exception in thread "main" java.lang.IllegalArgumentException: Response Survived not found in file file:/home/tamura/git/tribuo-examples/titanic/test_removed.csv
    at org.tribuo.data.csv.CSVLoader.validateResponseNames(CSVLoader.java:286)
    at org.tribuo.data.csv.CSVLoader.innerLoadFromCSV(CSVLoader.java:244)
    at org.tribuo.data.csv.CSVLoader.loadDataSource(CSVLoader.java:238)
    at org.tribuo.data.csv.CSVLoader.loadDataSource(CSVLoader.java:209)
    at org.tribuo.data.csv.CSVLoader.loadDataSource(CSVLoader.java:184)
    at org.tribuo.data.csv.CSVLoader.loadDataSource(CSVLoader.java:138)
    at TitanicSurvivalClassifier.main(TitanicSurvivalClassifier.java:74)
```

It's a little inconvenient.

**Solution**

Add methods to load data without response variable name.

Craigacp:
The `CSVLoader` is designed for purely numerical data, not the mixture of numerical and text data that comprises the Kaggle Titanic dataset, so this will probably error out later on anyway. For mixed numerical and text data you should use `CSVDataSource` and construct a `RowProcessor` which allows full control of how the features are extracted from the columnar data. `CSVDataSource` has a boolean `outputRequired` flag which if set to false will generate `Example`s which contain the unknown output for that output type (e.g. `Label.UNKNOWN` for classification tasks) if the output doesn't exist in the csv file.

However you're right that there should be a way to load a numerical CSV which doesn't have a response. We'll fix that.

k-tamura:
@Craigacp Thank you for your comment. I wrote the following code first:

```java
com.nhl.dflib.DataFrame df = com.nhl.dflib.csv.Csv.loader().load("titanic/train.csv");
System.out.println(df.head(5));
[...]

Issue #50, opened 2020-09-28
The way to select particular features from CSV file


galimru:
Hi, 

I'm new to ML stuff but decided to start with your library.

I have a question:
Given this dataset: https://archive.ics.uci.edu/ml/datasets/student+performance

I'm trying to make a linear regression to predict grade (G3) for students based on several features from this dataset.
I want to use to the following features: `G1, G2, studytime, failures, absences`.

How I can define which columns of the csv file need to use as features for dataset?
It seems to me the CSVLoader treats all columns which are not response names as features.
This leads to the exception because it's trying to convert all features to double:
```
Exception in thread "main" java.lang.NumberFormatException: For input string: "course"
	at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2043)
	at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
```

I've seen examples on python with Pandas library where you can pick the columns for features and another column for result.
Is it possible to do kind of like that with Tribuo?

Craigacp:
Yep, but to process csvs where you don't want to use all the columns then you need to use `RowProcessor` rather than the simple `CSVLoader`. If you instantiate a `CSVDataSource` you can pass it a `RowProcessor` instance which controls how columnar data is parsed. The `RowProcessor` accepts a `fieldProcessorMap` which gives the mapping between feature fields and the processor used to convert the String in that field into a list of Tribuo features, and a `ResponseProcessor` which you can use to pick the response column. The RowProcessor is very flexible so there are lots of other parameters but that should be sufficient to get you going.

galimru:
Great, this is very clear. I appreciate it.

Craigacp:
I'll add docs to the RowProcessor constructors in the docs pass this week.

spazbob:
Hi @Craigacp - I'm in the same boat, trying to get started with Tribuo and some basic CSV files with non-double Feature values. I've been working my way through the library but it would be great to have some code samples using RowProcessor etc. - had a quick look in the docs and tests but couldn't see anything obvious. Are there plans to build out the docs or have I just missed something?

Craigacp:
I'm writing a tutorial on the `RowProcessor` at the moment, it should land in a public branch sometime early next week.

PR #81, opened 2020-11-02
Lots of small documentation updates

### LINKED ISSUE #42: Include saving model in docs/tutorial
Can an example of serializing models and datasets to disk be included in the tutorials and/or docs? Additional help on using ONNX would be helpful too.

Craigacp:
### Description
Many small javadoc fixes, general docs fixes, tutorial updates, new tutorials on columnar data loading and external model imports.

In addition to docs fixes there are three bug fixes: `MultiLabelEvaluationImpl.toString()` had the same accuracy integer division bug as `LabelEvaluation.toFormattedString()`, `Event.equals` and `Event.hashCode` looked at the score when they are documented not to (though this is less relevant for those classes as the domains are specialised because there are only three possible values), and `DummyRegressionTrainer` was using an old provenance class not `TrainerProvenanceImpl` like it should be.

Finally it patches the pom file parent reference to fix an issue with project discovery.

### Motivation
Fixes for the docs found by our users after release.

Fixes #42, #50, #53, #60, #77.


AnomalyDetection/Core/src/main/java/org/tribuo/anomaly/AnomalyInfo.java
```diff
@@ -36,12 +36,30 @@
...
+     */
     protected long expectedCount = 0;
+
+    /**
+     * The number of anomalous events observed.
+     */
     protected long anomalyCount = 0;
+
+    /**
+     * The number of unknown events observed (i.e. those without labels).
```
eelstretching, line 50: Comma after i.e.

Classification/Core/src/main/java/org/tribuo/classification/baseline/DummyClassifierTrainer.java
```diff
@@ -39,7 +39,27 @@
  */
 public final class DummyClassifierTrainer implements Trainer<Label> {
```
eelstretching, line 40: Not super happy with Dummy? Maybe BaselineClassifierTrainer would be better? Not specifically introduced by this PR, but maybe we should fix it while we're here?

Common/XGBoost/src/main/java/org/tribuo/common/xgboost/XGBoostTrainer.java
```diff
@@ -63,7 +63,8 @@
  * </pre>
  * N.B.: This uses a native C implementation of xgboost that links to various C libraries, including libgomp
  * and glibc. If you're running on Alpine, which does not natively use glibc, you'll need to install glibc
- * into the container.
+ * into the container. On Windows this binary is not available in the Maven Central release, you'll need
+ * to compile it from source.
```
eelstretching, line 67: Do we have any extant advice on compiling XGBoost? Could we link to some tribuo.org page on how to do it? I guess we don't want to support the myriad Windows installs out there ("I couldn't get it to work on Windows ME!")

Core/src/main/java/org/tribuo/Example.java
```diff
@@ -47,8 +47,14 @@
...
 
+    /**
+     * The default initial size of the metadata map.
+     */
     protected static final int DEFAULT_METADATA_SIZE = 2;
 
+    /**
+     * The default weight.
+     */
     public static final float DEFAULT_WEIGHT = 1.0f;
```
eelstretching, line 58: What's the coding standard for numbers like 1 and 0. If I recall correctly, just using 1 and 0 will get you the 1 or 0 of the right type. Having it be explicit is probably better?

Data/src/main/java/org/tribuo/data/csv/package-info.java
```diff
@@ -17,5 +17,12 @@
 /**
  * Provides classes which can load columnar data (using a {@link org.tribuo.data.columnar.RowProcessor})
  * from a CSV (or other character delimited format) file.
+ * <p>
+ * {@link org.tribuo.data.csv.CSVDataSource} is the main way of loading CSV format data into Tribuo.
+ * It provides full control over featurisation, output processing and metadata extraction.
+ * {@link org.tribuo.data.csv.CSVLoader} is for simple numerical CSV files where all the
+ * non-response columns should be treated as features. {@link org.tribuo.data.csv.CSVSaver} writes
+ * out a * Tribuo {@link org.tribuo.Dataset} in CSV format suitable for loading via
```
eelstretching, line 25: What's that stray asterisk doing in there?

eelstretching (changes requested):
Address comments in the PR.

Classification/Core/src/main/java/org/tribuo/classification/baseline/DummyClassifierTrainer.java
Craigacp: It mirrors scikit-learn's names for these kind of classifiers/regressors. We can change it in the next major version bump if you want.

Common/XGBoost/src/main/java/org/tribuo/common/xgboost/XGBoostTrainer.java
Craigacp: Basically you need to install the Windows SDK to get the msbuild system and then run mvn package in the jvm-packages directory. I can talk to the XGBoost people about it and maybe firm up some instructions.

Core/src/main/java/org/tribuo/Example.java
Craigacp: I tend to put the specifier if it's a long, or if it's a float, and will put `d` on things if they are integers assigned to longs, but I usually prefer to put the decimal place in that case.

Data/src/main/java/org/tribuo/data/csv/package-info.java
Craigacp: Oops, merged the lines in vim. Fixed.

AnomalyDetection/Core/src/main/java/org/tribuo/anomaly/AnomalyInfo.java
Craigacp: Fixed.

eelstretching:
Looks good to me.

eelstretching (commented):
LGTM

eelstretching (approved):
LGTM

PR #82, opened 2020-11-03
Consistent Evaluator error messages

Craigacp:
### Description
`MultiLabelEvaluator` calls `MultiLabelConfusionMatrix` to tabulate the results, but if the ground truth labels contained the unknown label or a label that the model wasn't trained on then it gave a mysterious out of bounds exception from `DenseMatrix`. Now it consistently throws `IllegalArgumentException` with a useful error message saying what went wrong. A similar issue happened to clustering evaluator where it would silently treat the unassigned cluster id as a valid id and skew the metrics, and in regression evaluator which would propagate a NaN through in single dimensional regression and throw a variety of out of bounds exceptions in multidimensional regressions.

Also added some docs to `Evaluator` and subclasses to describe the kind of exception thrown.


jhalexand (changes requested):
It looks like this will only catch bad input in the ground truth once the predictions have already been made. It would be good to check earlier in the codepath to verify that the ground truth doesn't have unexpected values before the computation happens. We can head off some "Why didn't you tell me that earlier???" aggravation.

jhalexand (approved):
Looks good!

PR #85, opened 2020-11-10
Fixing a bug in the configuration tutorial example XML file.

Craigacp:
The example xml file for the tutorial has the wrong type for the MNIST test datasource. This PR fixes it.


eelstretching (approved):
Well, that's annoying. Looks good to me.

PR #90, opened 2020-11-18
Add CodeQL security scanning

daikikatsuragawa:
### Description

Add CodeQL security scanning. 

### Motivation

I propose to introduce CodeQL (code scanning) to this project.
This is to prevent security problems hidden in the code.

https://securitylab.github.com/tools/codeql

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,68 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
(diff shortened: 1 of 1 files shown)


Craigacp:
I've had a look through the default query set, and it flags a few things which aren't issues (like our use of hash functions for features). We also run static analysis tools internally on Tribuo, and I'm not sure that the default set of things for any tool is appropriate to turn on for all Tribuo PRs, as it's going to have a bunch of false positives that we need to disable (similar to the spotbugs excludes we already have). If there is a way of excluding CodeQL alerts which are definitely false positives then I'd be open to enabling this on the Tribuo repo. However before we can integrate any pull requests you'll need to sign the Oracle Contributor Agreement, following the process detailed in our [CONTRIBUTING.md](https://github.com/oracle/tribuo/blob/main/CONTRIBUTING.md) (or alternatively let me know the email address you've signed it under if you've already signed it).

If there is something specific that you found using a more detailed CodeQL analysis then please open an issue, or follow our security bug reporting guidelines (depending on the nature of the issue).

PR #91, opened 2020-11-18
Fixing a classification tree bug

Craigacp:
### Description
Fixing an issue where classification trees split incorrectly, adding a method to get the depth of a built tree for debugging, renamed a couple of methods.

### Motivation
Classification trees mirrored the left indices into the right indices, so the trees were broken and performed poorly.


jhalexand (approved):
Merging bugfix. Should add unit test in the future, though it'll be a fair bit of work (per Adam).

PR #97, opened 2020-11-26
Add an option to quiesce ONNX tests

Craigacp:
### Description
Adds a flag to the pom file which quiesces the ONNX tests.

### Motivation
The ONNX tests don't pass on Apple Silicon as I've only just added the support for it to the upstream release.

### DIFF
--- a/Interop/ONNX/pom.xml
+++ b/Interop/ONNX/pom.xml
@@ -100,6 +100,13 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


jhalexand (approved):
Makes sense. We should probably think if we want to remove the option once we have Apple Silicon support woven in.

Craigacp:
Yeah, though I've been considering making it a default part of the modules which use native code, as there will always be some CI environment they fail in.

jhalexand:
I actually edited my original comment to add “think if we want to” for exactly this reason.

Issue #72, opened 2020-10-22
Documentation: Proper DataSource format and usage for K-Means Clustering


lincolnthree:
**Is your feature request related to a problem? Please describe.**
Still a newbie to this library, so thanks for bearing with me.

Right now, the documentation shows how to run K-Means clustering on an auto-generated data set of Gaussian clusters. This is great, as it shows K-Means is possible, but (unless I'm missing something) it does not show the steps to input real data. (It mentions ` You can also use any of the standard data loaders to pull in clustering data.` but I don't see where that's documented).

I've figured out how to load a CSV file of features and metadata (thanks to your new Colunmar tutorial), but I can't seem to infer how to connect this data with `KMeansTrainer`, or if that's even the right approach.

**Describe the solution you'd like**
A clear and concise description/example of how to load real-world (non-autogenerated) data into the K-Means algorithm.

**Describe alternatives you've considered**
Looking through JavaDocs, but having trouble knowing what to focus on.

**Additional context**
![image](https://user-images.githubusercontent.com/362329/96885853-009b0d00-1451-11eb-9e3e-f2edc4d3a2ad.png)

lincolnthree:
I think I'm conflating `Label` with `ClusterId`, but looking for where in the docs I can clarify that.

lincolnthree:
If it helps, essentially what I'm trying to do is associate named features to coordinates such that they can be classified/clustered using K-Means. There's probably a way to do this already, but I'm missing it.

Craigacp:
Do you have ground truth cluster ids in your csv file? If not then you should add any response processor you like, make sure you pass it `ClusteringFactory` rather than `LabelFactory`, and then your datasource and dataset generic type should become `ClusterID` not `Label`. `Label` is only used for multiclass classification problems, `ClusterID` is the output type for clustering problems. Also make sure you pass the `false` flag to `CSVDataSource` so it knows it's ok for there not to be an output variable in the csv file.

If you do have ground truth cluster ids (and those ids are integers), then you should use `FieldResponseProcessor` as that will pass them through unaltered to the `ClusteringFactory` so they can be converted into `ClusterID` instances. If the ground truth clusters aren't integers, then at the moment you'll need to write a specific `ResponseProcessor` implementation that maps those values into the integers.

[...]

PR #98, opened 2020-11-26
Adds a better method for accessing the centroids of a KMeansModel

Craigacp:
### Description
Adds `KMeansModel.getCentroids` which returns a `List<List<Feature>>`. The old `getCentroidVectors` method requires the user to do the mapping from internal feature indices to feature names themselves, and the indices should never be required to use the library. I've updated the clustering tutorial to use the new method and also the K-Means++ initialisation added after the 4.0 release.

### Motivation
Adds the new centroid accessor. It also fixes a compatibility bug as `KMeansTrainer` had it's constructor changed in a backwards incompatible way, so a constructor with the correct signature was added.

This is an improvement discussed in #72.


jhalexand (approved):
Looks good.

PR #99, opened 2020-11-27
Adding an empty response processor that always returns an empty optional

Craigacp:
### Description
Adds a response processor which always returns an empty optional (i.e. no response). This is used by RowProcessors which operate on unlabelled data like clustering or anomaly detection tasks.

Also adds an additional constructor to `RowProcessor` to make it a little simpler to use.

### Motivation
Makes like simpler when loading clustering or anomaly detection columnar datasets.

Partial fix for #72.


jhalexand (approved):
Looks good

PR #104, opened 2020-12-08
Updates to improve the performance and logging of the transformer inf…

eelstretching:
…rastructure

### Description

The performance of the transformers was suboptimal when there was a large number of features. Got the transformation time for a large text classification problem down from more than two hours to less than two minutes. Along the way, I added some logging (at Level.FINE) so you can see what's happening, if you choose to do so.

As I was working on a text classification problem, I added and IDFTransformation that can set TFIDF weights in the example.

### Motivation

Transformer performance was far too slow in some cases.

### DIFF
--- a/Core/src/main/java/org/tribuo/impl/ArrayExample.java
+++ b/Core/src/main/java/org/tribuo/impl/ArrayExample.java
@@ -159,12 +159,19 @@ public class ArrayExample<T extends Output<T>> extends Example<T> {
      */
     public ArrayExample(Example<T> other) {
         super(other);
-        featureNames = new String[other.size()];
-        featureValues = new double[other.size()];
-        for (Feature f : other) {
-            featureNames[size] = f.getName();
-            featureValues[size] = f.getValue();
-            size++;
+        if(other instanceof ArrayExample) {
+            ArrayExample< T> otherArr = (ArrayExample<T>) other;
+            featureNames = Arrays.copyOf(otherArr.featureNames, otherArr.featureNames.length);
+            featureValues = Arrays.copyOf(otherArr.featureValues, otherArr.featureValues.length);
+            size = otherArr.size;
+        } else {
+            featureNames = new String[other.size()];
+            featureValues = new double[other.size()];
+            for(Feature f : other) {
+                featureNames[size] = f.getName();
+                featureValues[size] = f.getValue();
+                size++;
+            }
         }
     }
 
@@ -408,14 +415,33 @@ public class ArrayExample<T extends Output<T>> extends Example<T> {
 
     @Override
     public void transform(TransformerMap transformerMap) {
-        for (Map.Entry<String,List<Transformer>> e : transformerMap.entrySet()) {
-            int index = Arrays.binarySearch(featureNames,0,size,e.getKey());
-            if (index >= 0) {
-                double value = featureValues[index];
-                for (Transformer t : e.getValue()) {
+        if(transformerMap.size() < featureNames.length) {
+            //
+            // We have transformers than feature names, so let's
+            // iterate through the map and find the features.
+            for(Map.Entry<String, List<Transformer>> e : transformerMap.entrySet()) {
+                int index = Arrays.binarySearch(featureNames, 0, size, e.getKey());
+                if(index >= 0) {
+                    double value = featureValues[index];
+                    for(Transformer t : e.getValue()) {
+                        value = t.transform(value);
+                    }
+                    featureValues[index] = value;
+                }
+            }
+        } else {
+            //
+            // We have more transformers, so let's fetch them by name.
+            for(int i = 0; i < size; i++) {
+                List<Transformer> l = transformerMap.get(featureNames[i]);
+                if(l == null) {
+                    continue;
+                }
+                double value = featureValues[i];
+                for(Transformer t : l) {
                     value = t.transform(value);
                 }
-                featureValues[index] = value;
+                featureValues[i] = value;
             }
         }
     }
--- a/Core/src/main/java/org/tribuo/impl/ArrayExample.java
+++ b/Core/src/main/java/org/tribuo/impl/ArrayExample.java
@@ -161,8 +161,8 @@ public class ArrayExample<T extends Output<T>> extends Example<T> {
         super(other);
         if(other instanceof ArrayExample) {
             ArrayExample< T> otherArr = (ArrayExample<T>) other;
-            featureNames = Arrays.copyOf(otherArr.featureNames, otherArr.featureNames.length);
-            featureValues = Arrays.copyOf(otherArr.featureValues, otherArr.featureValues.length);
+            featureNames = Arrays.copyOf(otherArr.featureNames, otherArr.size);
+            featureValues = Arrays.copyOf(otherArr.featureValues, otherArr.size);
             size = otherArr.size;
         } else {
             featureNames = new String[other.size()];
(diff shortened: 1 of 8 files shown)


Core/src/main/java/org/tribuo/impl/ArrayExample.java
```diff
@@ -159,12 +159,19 @@ public ArrayExample(T output, List<? extends Feature> features) {
...
         super(other);
-        featureNames = new String[other.size()];
-        featureValues = new double[other.size()];
-        for (Feature f : other) {
-            featureNames[size] = f.getName();
-            featureValues[size] = f.getValue();
-            size++;
+        if(other instanceof ArrayExample) {
+            ArrayExample< T> otherArr = (ArrayExample<T>) other;
+            featureNames = Arrays.copyOf(otherArr.featureNames, otherArr.featureNames.length);
```
Craigacp, line 164: I know I wrote this, but the Arrays.copyOf should probably be to `size` not `otherArr.featureNames.length` to make it match up with the other side of the else. Plus it's slightly more memory efficient that way.

Core/src/main/java/org/tribuo/impl/ArrayExample.java
```diff
@@ -408,14 +415,33 @@ public void set(Feature feature) {
...
     @Override
     public void transform(TransformerMap transformerMap) {
-        for (Map.Entry<String,List<Transformer>> e : transformerMap.entrySet()) {
-            int index = Arrays.binarySearch(featureNames,0,size,e.getKey());
-            if (index >= 0) {
-                double value = featureValues[index];
-                for (Transformer t : e.getValue()) {
+        if(transformerMap.size() < featureNames.length) {
+            //
+            // We have transformers than feature names, so let's
```
Craigacp, line 420: "have transformers" -> "have fewer transformers"

Core/src/main/java/org/tribuo/impl/ArrayExample.java
```diff
@@ -408,14 +415,33 @@ public void set(Feature feature) {
...
+                    featureValues[index] = value;
+                }
+            }
+        } else {
+            //
+            // We have more transformers, so let's fetch them by name.
+            for(int i = 0; i < size; i++) {
+                List<Transformer> l = transformerMap.get(featureNames[i]);
+                if(l == null) {
+                    continue;
```
Craigacp, line 438: I think I'd prefer this test to be `if (l != null)` and that way we can get rid of the continue statement.

Core/src/main/java/org/tribuo/transform/TransformTrainer.java
```diff
@@ -81,10 +86,24 @@ public TransformTrainer(Trainer<T> innerTrainer, TransformationMap transformatio
 
     @Override
     public TransformedModel<T> train(Dataset<T> examples, Map<String, Provenance> instanceProvenance) {
+        
+        logger.info(String.format("%s %s", TransformTrainer.class.getName(), logger.getLevel()));
+        if(logger.isLoggable(Level.FINE)) {
+            logger.info(String.format("It says fine is loggable?"));
+        }
+        logger.fine(String.format("Creating transformers"));
+        logger.fine("Really doing it, though. Like for real");
```
Craigacp, line 95: I don't think this logging statement needs to be here.

Core/src/main/java/org/tribuo/transform/TransformTrainer.java
```diff
@@ -81,10 +86,24 @@ public TransformTrainer(Trainer<T> innerTrainer, TransformationMap transformatio
...
+        for(Handler h : logger.getHandlers()) {
+            h.flush();
```
Craigacp, line 97: Does this flush need to be here either?

Core/src/main/java/org/tribuo/transform/TransformTrainer.java
```diff
@@ -81,10 +86,24 @@ public TransformTrainer(Trainer<T> innerTrainer, TransformationMap transformatio
 
     @Override
     public TransformedModel<T> train(Dataset<T> examples, Map<String, Provenance> instanceProvenance) {
+        
+        logger.info(String.format("%s %s", TransformTrainer.class.getName(), logger.getLevel()));
+        if(logger.isLoggable(Level.FINE)) {
+            logger.info(String.format("It says fine is loggable?"));
```
Craigacp, line 92: Don't think this check needs to be here.

Core/src/main/java/org/tribuo/transform/TransformTrainer.java
```diff
@@ -29,6 +29,9 @@
 
 import java.time.OffsetDateTime;
 import java.util.Map;
+import java.util.logging.Handler;
```
Craigacp, line 32: Not sure we need the Handler import once the debugging statements have been removed.

Core/src/main/java/org/tribuo/transform/TransformationMap.java
```diff
@@ -49,9 +52,9 @@
 public class TransformationMap implements Configurable, Provenancable<ConfiguredObjectProvenance> {
 
     @Config(mandatory = true,description="Global transformations to apply after the feature specific transforms.")
-    private List<Transformation> globalTransformations;
+    private List<Transformation> globalTransformations = new ArrayList<>();
```
Craigacp, line 55: This doesn't need to be assigned to an ArrayList if it's mandatory. It'll be overwritten by the config system, and it's overwritten in the constructor (though the constructor could do a defensive copy I guess).

Core/src/test/java/org/tribuo/transform/SimpleTransformTest.java
```diff
@@ -232,7 +232,7 @@ public void testFeatureSpecific() {
         map.put("F6",Arrays.asList(SimpleTransform.exp(),SimpleTransform.log(),SimpleTransform.add(5)));
         map.put("F8",Arrays.asList(SimpleTransform.exp(),SimpleTransform.log()));
 
-        TransformationMap t = new TransformationMap(new ArrayList<>(),map);
+        TransformationMap t = new TransformationMap(new ArrayList<Transformation>(),map);
```
Craigacp, line 235: Not sure this change needs to be here.

Core/src/main/java/org/tribuo/transform/TransformerMap.java
```diff
@@ -129,16 +133,29 @@ public static double applyTransformerList(double value, List<Transformer> transf
...
         }
 
+        logger.fine(String.format("Transforming data set copy"));
+        
         newDataset.transform(this);
 
         return newDataset;
     }
+    
+    public int size() {
```
Craigacp, line 152: Javadoc for the new methods please.

Core/src/main/java/org/tribuo/transform/TransformerMap.java
```diff
@@ -129,16 +133,29 @@ public static double applyTransformerList(double value, List<Transformer> transf
...
+        return map.size();
+    }
+    
+    public List<Transformer> get(String featureName) {
```
Craigacp, line 156: Javadoc.

Craigacp (changes requested):
A few small things to fix.

Core/src/main/java/org/tribuo/impl/ArrayExample.java
eelstretching: I presume you mean otherArr.size here.

Core/src/main/java/org/tribuo/transform/TransformTrainer.java
eelstretching: There were actually a few that didn't need to be there, I took them all out, including the flush in the next one.

eelstretching:
Somehow I also managed to not add IDFTransformer to the original commit.

Core/src/main/java/org/tribuo/impl/ArrayExample.java
Craigacp: Yeah, that one.

Craigacp (approved):
LGTM

PR #106, opened 2020-12-10
Adds a multi-label linear sgd classifier

Craigacp:
### Description
This PR adds a native multi-label linear SGD model, refactors all the linear sgd models to share a base class, and adds a bunch of additional support to the math library for dense vectors (in preparation for further optimizations to the linear sgd models). It also switches over the vector normalizers to have an in place normalization method to reduce allocations in the inner loop of training and inference.

### Motivation
The multi-label package includes a wrapper to convert any classifier into a multi-label classifier, but this wrapper is too slow for large applications. The multi-label linear sgd classifier in this PR is an order of magnitude or so faster when ran on a large text corpus. It also preps the linear models package for further work introducing vectorisation, l2 regularisation and other improvements.


Classification/SGD/src/main/java/org/tribuo/classification/sgd/objectives/Hinge.java
```diff
@@ -52,14 +52,20 @@ public Hinge() {
...
     }
 
+    @Deprecated
+    @Override
+    public Pair<Double, SGDVector> valueAndGradient(int truth, SGDVector prediction) {
+        return lossAndGradient(truth, prediction);
+    }
+
     /**
      * Returns a {@link Pair} of {@link Double} and {@link SparseVector}.
```
eelstretching, line 62: The javadoc here seems a bit obvious? Feels like it should say the meaning of the Double and SparseVector you're returning? And for what purpose? Of course, that javadoc can be in the parent class, as we're overriding here.

Craigacp: Yeah, the interface has better javadoc. I didn't actually change it on this method, but I can tidy it up.

eelstretching (changes requested):
Just a few small changes here and there and it should be fine.

Common/SGD/src/main/java/org/tribuo/common/sgd/AbstractLinearSGDTrainer.java
```diff
@@ -0,0 +1,309 @@
...
+                for (int j = 0; j < sgdFeatures.length; j++) {
+                    SGDVector pred = linearParameters.predict(sgdFeatures[j]);
+                    Pair<Double,SGDVector> output = objective.lossAndGradient(sgdTargets[j],pred);
+                    loss += output.getA()*weights[j];
+
+                    Tensor[] updates = localOptimiser.step(linearParameters.gradients(output,sgdFeatures[j]),weights[j]);
+                    linearParameters.update(updates);
+
+                    iteration++;
+                    if ((iteration % loggingInterval == 0) && (loggingInterval != -1)) {
```
eelstretching, line 199: This test should be the other way around, since we're more likely to not be logging (and maybe it should be < 0). It's in that order in the minibatch > 1 case.

Math/src/main/java/org/tribuo/math/la/DenseMatrix.java
```diff
@@ -47,6 +52,10 @@ public DenseMatrix(int dim1, int dim2) {
         this.numElements = dim1*dim2;
     }
 
+    /**
+     * Copies the supplied matrix.
+     * @param other The matrix to copy.
+     */
```
eelstretching, line 58: Just curious why we're not using arraycopy in this constructor?

MultiLabel/SGD/src/main/java/org/tribuo/multilabel/sgd/linear/LinearSGDTrainer.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param minibatchSize The size of any minibatches.
+     * @param seed A seed for the random number generator, used to shuffle the examples before each epoch.
+     */
+    public LinearSGDTrainer(MultiLabelObjective objective, StochasticGradientOptimiser optimiser, int epochs, int loggingInterval, int minibatchSize, long seed) {
+        super(optimiser,epochs,loggingInterval,minibatchSize,seed);
+        this.objective = objective;
+    }
+
+    /**
+     * Sets the minibatch size to 1.
```
eelstretching, line 66: This comment actually doesn't describe what the method is. I saw this in a couple of the other versions of the Trainer constructors too.

Common/SGD/src/main/java/org/tribuo/common/sgd/AbstractLinearSGDTrainer.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    protected abstract String getModelClassName();
+
+    @Override
+    public TrainerProvenance getProvenance() {
+        return new TrainerProvenanceImpl(this);
+    }
+
+    /**
+     * In place shuffle of the features, outputs and weights.
```
eelstretching, line 283: Javadoc comments for methods should have the first sentence start with a verb.

eelstretching (changes requested):
See the attached changes.

Math/src/main/java/org/tribuo/math/la/DenseMatrix.java
Craigacp: Because it's not final and `ShrinkingMatrix` and `AdaGradRDAMatrix` both subclass `DenseMatrix` and mess with the get method to apply a transformation as part of the regularisation during training. I guess I could check if it's only a DenseMatrix without any other classes and then do an arraycopy, falling back to this code otherwise, but it doesn't seem worth it at the moment.

Classification/SGD/src/main/java/org/tribuo/classification/sgd/objectives/Hinge.java
Craigacp: Fixed.

Craigacp:
@eelstretching any thoughts on using the Pair vs the concrete class to return the gradient and the loss?

eelstretching (dismissed):
The changes look good. I think returning the Pair is the correct decision.

Craigacp:
I removed the `LossAndGradient` class as we've decided on keeping the pair.

eelstretching (approved):
LGTM!

PR #107, opened 2021-01-22
Upgrades XGBoost to 1.3.1

Craigacp:
### Description
Bumps XGBoost from 1.0.0 to 1.3.2. Also plumbs through additional XGBoost parameters, so now the booster type and tree method are controllable. Exposes the logging verbosity as a replacement for the deprecated silent parameter.

### DIFF
--- a/Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
+++ b/Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
@@ -38,7 +38,7 @@ public class XGBoostOptions implements ClassificationOptions<XGBoostClassificati
     public int xgbEnsembleSize = -1;
     @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights.")
     public float xbgAlpha = 0.0f;
-    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (range [0,inf]).")
+    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (range [0,Infinity]).")
     public float xgbMinWeight = 1;
     @Option(longName = "xgb-max-depth", usage = "Max tree depth (range (0,Integer.MAX_VALUE]).")
     public int xgbMaxDepth = 6;
@@ -46,7 +46,7 @@ public class XGBoostOptions implements ClassificationOptions<XGBoostClassificati
     public float xgbEta = 0.3f;
     @Option(longName = "xgb-subsample-features", usage = "Subsample features for each tree (range (0,1]).")
     public float xgbSubsampleFeatures = 0.0f;
-    @Option(longName = "xgb-gamma", usage = "Minimum loss reduction to make a split (range [0,inf]).")
+    @Option(longName = "xgb-gamma", usage = "Minimum loss reduction to make a split (range [0,Infinity]).")
     public float xgbGamma = 0.0f;
     @Option(longName = "xgb-lambda", usage = "L2 regularization term for weights.")
     public float xgbLambda = 1.0f;
--- a/Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
+++ b/Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
@@ -19,32 +19,44 @@ package org.tribuo.classification.xgboost;
 import com.oracle.labs.mlrg.olcut.config.Option;
 import org.tribuo.Trainer;
 import org.tribuo.classification.ClassificationOptions;
+import org.tribuo.common.xgboost.XGBoostTrainer;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * CLI options for training an XGBoost classifier.
  */
 public class XGBoostOptions implements ClassificationOptions<XGBoostClassificationTrainer> {
+    private static final Logger logger = Logger.getLogger(XGBoostOptions.class.getName());
+
+    @Option(longName = "xgb-booster-type", usage = "Weak learning algorithm.")
+    public XGBoostTrainer.BoosterType xgbBoosterType = XGBoostTrainer.BoosterType.GBTREE;
+    @Option(longName = "xgb-tree-method", usage = "Tree building algorithm.")
+    public XGBoostTrainer.TreeMethod xgbTreeMethod = XGBoostTrainer.TreeMethod.AUTO;
     @Option(longName = "xgb-ensemble-size", usage = "Number of trees in the ensemble.")
     public int xgbEnsembleSize = -1;
-    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights (default 0).")
+    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights.")
     public float xbgAlpha = 0.0f;
-    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (default 1, range [0,inf]).")
+    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (range [0,inf]).")
     public float xgbMinWeight = 1;
-    @Option(longName = "xgb-max-depth", usage = "Max tree depth (default 6, range (0,inf]).")
+    @Option(longName = "xgb-max-depth", usage = "Max tree depth (range (0,inf]).")
     public int xgbMaxDepth = 6;
-    @Option(longName = "xgb-eta", usage = "Step size shrinkage parameter (default 0.3, range [0,1]).")
+    @Option(longName = "xgb-eta", usage = "Step size shrinkage parameter (range [0,1]).")
     public float xgbEta = 0.3f;
-    @Option(longName = "xgb-subsample-features", usage = "Subsample features for each tree (default 1, range (0,1]).")
+    @Option(longName = "xgb-subsample-features", usage = "Subsample features for each tree (range (0,1]).")
     public float xgbSubsampleFeatures;
-    @Option(longName = "xgb-gamma", usage = "Minimum loss reduction to make a split (default 0, range [0,inf]).")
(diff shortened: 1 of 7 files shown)


Craigacp:
I rolled this back to 1.3.1 as that version has a macOS binary in Maven Central, and the fixes in 1.3.2 are Python or Solaris related.

nezda:
I'm trying this out locally on macOS (11.2.1). Seems to work fine. With `silent` `false` I see a full path on someone else's machine which seems odd:

```
[13:29:22] INFO: /Users/nanzhu/code/xgboost/src/tree/updater_prune.cc:101: tree pruning end, 24 extra nodes, 0 pruned nodes, max_depth=6
```

Also multi-threading doesn't seem to be working (just based on looking at CPU usage - tried `numThreads` 8), however it did train 5x faster than the previous version! The training time (and CPU usage) didn't seem to be impacted by the `numThreads` parameter.

Craigacp:
I believe the JVM builds of XGBoost for macOS are currently compiled by the developers rather than the CI, and the logger gives the line of the code and it's path for each logging message. So I think thats probably compiled by https://github.com/CodingCat who does a lot of the JVM work in XGBoost. This is probably changing in the next few releases, I've been talking to them about adding Windows support to the Maven Central builds - https://github.com/dmlc/xgboost/pull/6630#issuecomment-790702133. If that lands in the upcoming 1.4.0 release (and that release happens before Tribuo's 4.1 release) we'll update to that version to make things easier for our users.

I'm a little worried about the multithreading aspect, I'll see if I can replicate it.

Craigacp:
Ah, so I'd missed the threading issue because internally we build XGBoost4j with OpenMP turned on for Windows, macOS and Linux, but it's not turned on in the builds in Maven Central for macOS. I'll open an issue upstream as it looks like they build the Python macOS whl with OpenMP turned on.

Craigacp:
On Linux with this branch building a 500 tree model on MNIST (using Tribuo's default parameters) takes 30s with 1 thread and 10s with 6 threads (on an Intel Core i7-8700), so the Tribuo side of things is definitely passing down the right parameters.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
```diff
@@ -19,32 +19,44 @@
...
+    public XGBoostTrainer.BoosterType xgbBoosterType = XGBoostTrainer.BoosterType.GBTREE;
+    @Option(longName = "xgb-tree-method", usage = "Tree building algorithm.")
+    public XGBoostTrainer.TreeMethod xgbTreeMethod = XGBoostTrainer.TreeMethod.AUTO;
     @Option(longName = "xgb-ensemble-size", usage = "Number of trees in the ensemble.")
     public int xgbEnsembleSize = -1;
-    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights (default 0).")
+    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights.")
     public float xbgAlpha = 0.0f;
-    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (default 1, range [0,inf]).")
+    @Option(longName = "xgb-min-weight", usage = "Minimum sum of instance weights needed in a leaf (range [0,inf]).")
```
JackSullivan, line 41: How does one specify inf here? `Float.MAX_VALUE`? That maybe should be documented.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
```diff
@@ -19,32 +19,44 @@
...
+    private static final Logger logger = Logger.getLogger(XGBoostOptions.class.getName());
+
+    @Option(longName = "xgb-booster-type", usage = "Weak learning algorithm.")
+    public XGBoostTrainer.BoosterType xgbBoosterType = XGBoostTrainer.BoosterType.GBTREE;
+    @Option(longName = "xgb-tree-method", usage = "Tree building algorithm.")
+    public XGBoostTrainer.TreeMethod xgbTreeMethod = XGBoostTrainer.TreeMethod.AUTO;
     @Option(longName = "xgb-ensemble-size", usage = "Number of trees in the ensemble.")
     public int xgbEnsembleSize = -1;
-    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights (default 0).")
+    @Option(longName = "xgb-alpha", usage = "L1 regularization term for weights.")
```
JackSullivan, line 39: Why are we removing the defaults here?

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
```diff
@@ -19,32 +19,44 @@
...
     public float xgbGamma = 0.0f;
-    @Option(longName = "xgb-lambda", usage = "L2 regularization term for weights (default 1).")
+    @Option(longName = "xgb-lambda", usage = "L2 regularization term for weights.")
     public float xgbLambda = 1.0f;
-    @Option(longName = "xgb-quiet", usage = "Make the XGBoost training procedure quiet.")
+    @Option(longName = "xgb-quiet", usage = "Deprecated, use xgb-loglevel.")
     public boolean xgbQuiet;
-    @Option(longName = "xgb-subsample", usage = "Subsample size for each tree (default 1, range (0,1]).")
+    @Option(longName = "xgb-loglevel", usage = "Make the XGBoost training procedure quiet.")
+    public XGBoostTrainer.LoggingVerbosity xgbLogLevel = XGBoostTrainer.LoggingVerbosity.WARNING;
```
JackSullivan, line 56: Is there an reason to prefer a custom enum here to `java.util.logging.Level`?

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
```diff
@@ -19,32 +19,44 @@
...
+    @Option(longName = "xgb-subsample", usage = "Subsample size for each tree (range (0,1]).")
     public float xgbSubsample = 1.0f;
-    @Option(longName = "xgb-num-threads", usage = "Number of threads to use (default 4, range (1, num hw threads)).")
+    @Option(longName = "xgb-num-threads", usage = "Number of threads to use (range (1, num hw threads)).")
```
JackSullivan, line 59: Doesn't this need a default value?

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
Craigacp: OLCUT auto generates the defaults in the usage string by inspecting the fields now.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
Craigacp: There aren't as many levels of xgboost logger, as it's a value passed directly into the C library, so I figure rather than having a potentially confusing mapping I'd just expose what they have.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
Craigacp: OLCUT auto-generates it.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
Craigacp: Well if you do specify inf I don't think it can make any splits, so it probably goes pop. I should check the XGBoost docs again, this comes from their CLI docs but I don't think that line has been updated since XGBoost 0.7 ish.

JackSullivan (commented):
This looks good overall, there are a couple of UI nitpicks in `XGBoostOptions` I'd like clarification on. See inline comments.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
JackSullivan: Sorry, I meant doesn't the actual variable need a default value. Or at least, shouldn't it be required? Will xgboost crash if you give it zero threads or use all hardware threads?

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
JackSullivan: Fair enough

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
JackSullivan: I guess this same questions works better on `xgb-max-depth`, which has the same range.

Classification/XGBoost/src/main/java/org/tribuo/classification/xgboost/XGBoostOptions.java
Craigacp: If you give it zero it uses all hardware threads. I should note that in the usage.

JackSullivan (approved):
Looks good to me

nezda:
> Ah, so I'd missed the threading issue because internally we build XGBoost4j with OpenMP turned on for Windows, macOS and Linux, but it's not turned on in the builds provided by dmlc in Maven Central for macOS. I'll open an issue upstream as it looks like they build the Python macOS whl with OpenMP turned on.

I got multi-threading working on macOS following https://xgboost.readthedocs.io/en/latest/jvm/index.html#enabling-openmp-for-mac-os directions and building that and this project with `<xgboost.version>1.4.0-SNAPSHOT</xgboost.version>` - quite the time saver 👍

Craigacp:
Excellent. Looks like the xgboost developers will discuss turning it on in xgboost builds after the upcoming 1.4.0 release. I'm hopeful that that 1.4.0 release will include windows binaries though which will make things much simpler for Tribuo.

PR #112, opened 2021-02-06
Add dense feature vector optimisations to CRF and LinearSGD models

Craigacp:
### Description
This PR adds support for dense feature vectors to the CRF trainer/model and the linear SGD trainer/model. The CRF & linear model only use dense feature vectors if the example is truly dense (i.e. it contains a feature for each entry in the model or dataset's feature map). At some point in the future we could make this a model/trainer parameter, as the dense representation is probably faster at the cost of additional memory in spaces which are moderately sparse. We should do some benchmarking to find a reasonable point to make the default if we do this change.

To support the CRF and linear model change then there were additional methods implemented in the `org.tribuo.math.la` package, along with extra cases inside those methods. Additionally the `ModelProvenance` now correctly tracks if the dataset was dense, and the dataset better understands if it is dense (rather than only assuming it is dense if it had `densify` called on it). Only `MutableDataset` and `MutableSequenceDataset` know if they are dense or not, we should probably move this check onto `Dataset` and `SequenceDataset` so immutable views can be considered dense, but the nature of the views mean this is a little more complicated to check.

Several classes were added to various sequence packages to build them out for further testing. The main ones are an `IndependentSequenceTrainer` and an `IndependentSequenceModel` which wrap a regular `Trainer`/`Model` and use it to make independent predictions for each element of the sequence. There is also a more useful sequence train test harness for comparing different sequence models (and a small fix for `ViterbiTrainer` which was not compatible with the configuration system).

### Motivation
Working with a dense feature space (e.g. an embedding vector space) is many times faster than treating the dense space as if it is sparse with the associated indirection on all the linear algebra operations. This greatly improves the speed of the CRF operating over embedding vectors, and of the linear model when operating on dense spaces. The output of the two systems is identical.

This density optimisation opens the path for further speedups either by making Tribuo's la package more amenable to autovectorisation, or by explicitly vectorising it using the Vector API in Java 16.

### DIFF
--- a/Classification/Core/src/main/java/org/tribuo/classification/sequence/SeqTrainTest.java
+++ b/Classification/Core/src/main/java/org/tribuo/classification/sequence/SeqTrainTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2015-2020, Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.tribuo.classification.sequence;
+
+import com.oracle.labs.mlrg.olcut.config.ConfigurationManager;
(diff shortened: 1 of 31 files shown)


eelstretching (approved):
Looks good to me. More of a comment than a question: we're clearly planning to densify or sparsify based on characteristics of the examples vs. the feature map. I believe that we should also have a way to let the users pick what they want to do, even if they're going to pick a dumb way.

This will be more important when we get to the case where dense is not just "uses all the features"

Craigacp:
I agree there should be a user parameter. I'm thinking a flag on the trainer and model with an associated threshold float between zero and one. It won't be useful on most of the models as the backends are also sparse, but LinearSGDModel and CRFModel are likely targets.

Craigacp:
But that can be something for the next release.

PR #113, opened 2021-02-08
Adds support for standardising regression values in LibSVM

### LINKED ISSUE #73: Scaling/Rescaling needed for regression outputs
**Is your feature request related to a problem? Please describe.**
Currently scaling of training data (via transformations such as MeanStdDevTransformation) only applies to the features but not to the outputs; however regression outputs need to be scaled as well for some models to train properly. Specifically, this causes the RBF SVM to perform much more poorly comparing to scikit-learn, where it's easy to scale the entire dataset.

**Describe the solution you'd like**
Adding the option to scale the output of a training dataset in addition to the features when training a regressor. This also means that the output of the regressor will be inverse-scaled when performing predictions.

**Describe alternatives you've considered**
'Manually' scaling and inverse-scaling outside Tribuo's training/prediction flow. This is cumbersome, and in addition will not be included in the provenance.

**Additional context**

Craigacp:
### Description
Adds an option to LibSVMRegressionTrainer which standardises the input (i.e. ensures it's mean zero variance one) before training a model. It then applies the inverse transformation to the predictions to map them back into the correct output space. It also adds a non-linear regression data generator for testing, adds a couple of related methods in `org.tribuo.util.Util` and tidies up some related comments.

The internal implementation is quite ugly as it extends `svm_model` to poke in the mean and variance, because the types in the base `LibSVMTrainer` are too restrictive. When we have a breaking version it might be worth refactoring this and `LibLinearTrainer` (which is similar) to make the internal types controlled by Tribuo so they can be more easily extended if necessary (or just make them `Map<String,Object>`).

The standardization is off by default to preserve compatibility with Tribuo 4.0, but the performance is pretty poor on non-linear problems with it turned off, so it might be better to turn it on by default.

### Motivation
LibSVM has a real problem with non-standardized regression inputs, and performs very poorly. Using the new non-linear data generator an RBF SVM gets:
```
Multi-dimensional Regression Evaluation
RMSE = {Y=24.412388852859543}
Mean Absolute Error = {Y=11.827119549115682}
R^2 = {Y=0.7729975961406879}
explained variance = {Y=0.773032506305136}
```
and with standardization it gets:
```
Multi-dimensional Regression Evaluation
RMSE = {Y=1.214990405958973}
Mean Absolute Error = {Y=0.9604027608577747}
R^2 = {Y=0.9994377161686859}
explained variance = {Y=0.9994384243963165}
```

Fixes #73. I tested the other regressors and none of them exhibit the same issue on non-standardized data. If there are future reports of a similar issue in the other regressors we can roll out a similar solution.

### DIFF
--- a/Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
+++ b/Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.tribuo.util;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * An accumulator for online calculation of the mean and variance of a
+ * stream of doubles.
+ * <p>
+ * Note this class is not thread safe.
+ */
+public final class MeanVarianceAccumulator implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private double max = Double.NEGATIVE_INFINITY;
+    private double min = Double.POSITIVE_INFINITY;
+
+    private double mean = 0.0;
+    private double sumSquares = 0.0;
+
+    private long count = 0;
+
+    /**
+     * Constructs an empty mean/variance accumulator.
+     */
+    public MeanVarianceAccumulator() {}
+
+    /**
+     * Constructs a mean/variance accumulator and observes the supplied array.
+     * @param values The array to operate on.
+     */
+    public MeanVarianceAccumulator(double[] values) {
+        observe(values);
+    }
+
+    /**
+     * Copy constructor.
+     * @param other The MeanVarianceAccumulator to copy.
+     */
+    public MeanVarianceAccumulator(MeanVarianceAccumulator other) {
+        this.max = other.max;
+        this.min = other.min;
+        this.mean = other.mean;
+        this.sumSquares = other.sumSquares;
+        this.count = other.count;
+    }
+
+    /**
+     * Resets this accumulator to the starting state.
+     */
+    public void reset() {
+        this.max = Double.NEGATIVE_INFINITY;
+        this.min = Double.POSITIVE_INFINITY;
+        this.mean = 0;
+        this.sumSquares = 0;
+        this.count = 0;
+    }
+
+    /**
+     * Observes a value, i.e. updates the mean, variance, max and min statistics.
+     * @param value The value to observe.
+     */
+    public void observe(double value) {
+        if (value < min) {
+            min = value;
+        }
+        if (value > max) {
+            max = value;
+        }
+        count++;
+        double delta = value - mean;
+        mean += delta / count;
+        double delta2 = value - mean;
+        sumSquares += delta * delta2;
+    }
+
+    /**
+     * Observes an array of values, i.e. updates the mean, variance, max and min statistics.
+     * @param values The values to observe.
+     */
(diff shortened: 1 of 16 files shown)


Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
```diff
@@ -0,0 +1,171 @@
...
+    public void reset() {
+        this.max = Double.NEGATIVE_INFINITY;
+        this.min = Double.POSITIVE_INFINITY;
+        this.mean = 0;
+        this.sumSquares = 0;
+        this.count = 0;
+    }
+
+    /**
+     * Observes a value, i.e. updates the mean, variance, max and min statistics.
```
eelstretching, line 75: I get what you're going for here, but this comment is actually inaccurate. It doesn't update the variance, because that's computed when it's asked for.

Core/src/main/java/org/tribuo/util/Util.java
```diff
@@ -1040,4 +1055,36 @@ public static String formatDuration(long startMillis, long stopMillis) {
         return diffIndicesList.stream().mapToInt(Integer::intValue).toArray();
     }
 
+    /**
+     * Standardizes the input, i.e. divides it by the variance and subtracts the mean.
```
eelstretching, line 1059: The i.e. comment here is weird. Should say something like "a value is standardized by subtracting the mean and dividing the result by the variance". The way it's stated sounds like we divide first and then subtract, which is incorrect (assuming the code is correct, which I think it is.)

standardizeInPlace javadoc has the same problem.

I guess I'm curious why these methods are static when we have class members for doing the accumulation? Are they used elsewhere? If they are and staticness makes sense for these, then there should be non-static ones that use the mean and variance that we've accumulated on an input array.

Regression/Core/src/main/java/org/tribuo/regression/example/GaussianDataSource.java
```diff
@@ -54,10 +54,10 @@
     private int numSamples;
 
     @Config
-    private float slope;
+    private float slope = 0.0f;
```
eelstretching, line 57: 0 always has the right type, unless I miss my guess. 0.0f seems like gilding the lily here.

Regression/Core/src/main/java/org/tribuo/regression/example/NonlinearGaussianDataSource.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * Generates a single dimensional output drawn from
+ * N(w_0*x_0 - w_1*x_1 + w_2*x_1*x_0 + w_3*x_1*x_1*x_1 + intercept,variance).
```
eelstretching, line 47: Pretty sure the - w_1*x_1 should be + in the javadoc here, to match the code below? This error appears elsewhere in the javadoc.

eelstretching (commented):
Looks good, mostly javadoc fixes.

Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
Craigacp: Fixed.

Core/src/main/java/org/tribuo/util/Util.java
Craigacp: I added class members to the `MeanVarianceAccumulator` which performs standardization by calling into `Util.standardize`. The `MeanVarianceAccumulator` didn't end up being used by the standardization code in `LibSVMRegressionTrainer`, but my plan is to migrate all other the places where I do this over to this code over time, so I put it into this PR along with it's test.

Regression/Core/src/main/java/org/tribuo/regression/example/GaussianDataSource.java
Craigacp: My IDE complains and it doesn't compile without it.

Craigacp: I also added description fields to all the config variables while I was in here.

Regression/Core/src/main/java/org/tribuo/regression/example/NonlinearGaussianDataSource.java
Craigacp: I fixed the javadoc & comment. When I realised I'd made weight parameterisable the sign didn't matter anymore and I forgot to update the docs.

eelstretching (approved):
Looks good!

PR #114, opened 2021-02-11
Liblinear upgrade to 2.42, adds one-class SVM for anomaly detection

Craigacp:
### Description
Upgrades liblinear-java to 2.42 from 2.30. That new version adds one-class SVM, and this is exposed up as an anomaly detection algorithm in Tribuo.

### Motivation
Keeps current with Tribuo's dependencies, and expands the set of anomaly detection algorithms (though this one-class SVM is similar to the one provided by LibSVM, just faster).


AnomalyDetection/LibLinear/src/main/java/org/tribuo/anomaly/liblinear/LibLinearAnomalyTrainer.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+public class LibLinearAnomalyTrainer extends LibLinearTrainer<Event> {
+
+    private static final Logger logger = Logger.getLogger(LibLinearAnomalyTrainer.class.getName());
+
+    @Config(description = "Nu parameter in one class SVM.")
+    private double nu = 0.5;
+
+    /**
+     * Creates a trainer using the default values (ONECLASS_SVM, 1, 0.1, 0.5).
```
jhalexand, line 66: The JavaDoc doesn't match the implementation here (for maxIterations) or in the next few constructors

AnomalyDetection/LibLinear/configs/liblinear-example.xml
```diff
@@ -0,0 +1,37 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!--
+    Description:
+        An example configuration file for a one-class SVM optimized in the dual.
```
jhalexand, line 21: Since we're editing comments anyway, I'm not sure what "optimized in the dual" is supposed to mean here.

jhalexand (commented):
Code all looks good.  Couple comments about the doc.

AnomalyDetection/LibLinear/src/main/java/org/tribuo/anomaly/liblinear/LibLinearAnomalyTrainer.java
Craigacp: Ooops, yeah I should put max iterations in.

AnomalyDetection/LibLinear/configs/liblinear-example.xml
Craigacp: There are two ways of solving the linear SVM problem, the primal form which has variables per feature and the dual form which has variables per data point. The nu-SVM here only works in the dual form. The comment is probably irrelevant as it doesn't really add much information.

jhalexand (approved):
Looks good to me

PR #115, opened 2021-02-12
Provenance updates

Craigacp:
### Description
Fixes a provenance marshalling bug in `ModelProvenance` and `EnsembleModelProvenance` and adds a test to ensure it doesn't happen again.

Also adds system information to `ModelProvenance` and `EnsembleModelProvenance`. Now these provenances track "java.version", "os.name" and "os.arch" from the Java properties. They can be removed by passing the appropriate flag on construction, or by stripping them using `StripProvenance` (and so StripProvenance gained the ability to remove system provenances). 

### Motivation
Tribuo 4.0.2 can produce different outputs on Java 8 x86_64, Java 9+ x86_64 and Java on aarch64, due to differences in the implementation of Math.exp, Math.log and Math.sqrt on those platforms. To ensure replicability we need to track what platform the model was created on until we've run it down and either switched to the slower StrictMath for correctness or learned to live with it. Future optimisations using the Vector API will also likely induce model divergence based on the hardware as the order of operations will be different when training models with SGD, so we might need this permanently.


jhalexand (approved):
Looks good

PR #116, opened 2021-03-02
Adds a BERT feature extractor

Craigacp:
### Description
Adds a BERTFeatureExtractor which implements `TextPipeline` and `TextFeatureExtractor` so it can be used in document classification, and also methods which extract `SequenceExample` so it can be used in sequence prediction tasks. At the moment the sequence prediction bits don't implement an interface as there is no comparable sequence prediction text interface. We might consider introducing one later, and this class will be retrofitted to implement such an interface.

It loads BERT models which have been exported as onnx files from [HuggingFace Transformers](https://github.com/huggingface/transformers), and also loads in a trained wordpiece tokenizer from HuggingFace's tokenizer json format. This latter part is why we added a reference to HuggingFace Transformers in THIRD_PARTY_LICENSES.txt on the advice of legal even though we don't use any HuggingFace code.

At the moment passing in a String to the document embedding extractors will truncate the string to `maxLength` wordpieces, however the methods which pass in a pre-tokenized input (of the kind that is used in sequence prediction) don't check the length. We could consider adding such a check and throwing an exception if it's above `maxLength`.

### Motivation
BERT is very popular as a feature extractor for NLP tasks, but it's very tricky to use from Java or Tribuo. This makes it straightforward. Note it's already possible to use a BERT model which makes predictions by loading it into `ONNXExternalModel`, this PR is purely to allow it to be used as a feature extractor where the predictor is trained in Tribuo.

### Paper reference
[BERT: Pre-training of Deep Bidirectional Transformers for Language Understanding](https://www.aclweb.org/anthology/N19-1423/)


Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
```diff
@@ -0,0 +1,988 @@
...
+        List<String> tokens = tokenize(data);
+        return extractExample(tokens,output);
+    }
+
+    /**
+     * Tokenizes the input using the loaded tokenizer, truncates the
+     * token list if it's longer than {@code maxLength} - 2 (to account
+     * for [CLS] and [SEP] tokens), and then passes the token
+     * list to {@link #extractExample}.
+     * @param data The input text.
```
k8si, line 855: Maybe document what "tag" is supposed to be? 

Also, maybe add another method `process(String sequence1, String sequence2)` to support sentence pair tasks? (Same with the "extract" method)

Interop/ONNX/src/test/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractorTest.java
```diff
@@ -0,0 +1,136 @@
...
+                hidden_size=50,
+                num_hidden_layers=1,
+                num_attention_heads=1,
+                intermediate_size=200,
+                max_position_embeddings=100
+        )
+        */
+        Path modelPath = Paths.get(BERTFeatureExtractorTest.class.getResource("tinybert.onnx").toURI());
+        LabelFactory factory = new LabelFactory();
+        String input = "It is a truth universally acknowledged,";
```
k8si, line 83: I'd maybe add a couple more tests to check out-of-vocab behavior (e.g. Arabic or CJK characters, emoji) just so we know what happens. I'm not sure what's OOV for the full bert-base-cased tokenizer but should be easy to check with tinybert.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
```diff
@@ -0,0 +1,988 @@
...
+    /**
+     * Passes the tokens through BERT, replacing any unknown tokens with the [UNK] token.
+     * <p>
+     * The features of the returned example are dense, and come from the [CLS] token.
+     * <p>
+     * Throws {@link IllegalStateException} if the BERT model failed to produce an output.
+     * @param tokens The input tokens. Should be tokenized using the Tokenizer this BERT expects.
+     * @return A dense example representing the pooled output from BERT for the input tokens.
+     */
+    public Example<T> extractExample(List<String> tokens) {
```
k8si, line 617: Is there a check somewhere along this path to make sure `tokens` doesn't exceed max length?

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
```diff
@@ -0,0 +1,988 @@
...
+        return tokens;
+    }
+
+    /**
+     * Runs BERT on the input, returning the tokens, ids, masks and embeddings.
+     * @param data The input text.
+     * @return The tokens, the token ids, the token types, the masks, the cls embedding and the token embeddings.
+     * @throws OrtException If the native runtime failed.
+     */
+    private BERTResult bert(String data) throws OrtException {
```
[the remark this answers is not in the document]
Craigacp: No, that entry point is mostly for testing. In general Tribuo doesn't do batching in the feature processing stage so there's no support for it here.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
```diff
@@ -0,0 +1,988 @@
...
+     * Tokenizes the input using the loaded tokenizer, truncates the
+     * token list if it's longer than {@code maxLength} - 2 (to account
+     * for [CLS] and [SEP] tokens).
+     * @param data The input text.
+     * @return The wordpiece tokens for the supplied data.
+     */
+    List<String> tokenize(String data) {
+        List<String> tokens = tokenizer.split(data);
+        if (tokens.size() > (maxLength - 2)) {
+            logger.info("Truncating sentence to " + (maxLength + 2) + " from " + tokens.size());
```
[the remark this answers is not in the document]
Craigacp: Sure.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
Craigacp: These methods are constrained by Tribuo's text processing API which doesn't have the notion of sentence pairs. When we add structured inputs then there might be API endpoints for that structure, but we've not finished designing that.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
Craigacp: Nope, not on this path. I wasn't sure if truncating it or throwing an `IllegalArgumentException` was better. This is supposed to be pretokenized input.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
Craigacp: Ooops, yeah I was missing the `@param`. Thanks.

Interop/ONNX/src/test/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractorTest.java
Craigacp: There are tests for out-of-vocab wordpiece tokenizer behaviour in the tokens module.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/extractors/BERTFeatureExtractor.java
Craigacp: I added `IllegalArgumentException` to the methods that accept a list as the programmer could check if it's longer than `getMaxLength` and deal with it themselves.

k8si (approved):

PR #122, opened 2021-03-27
Fixing the transformer behaviour on sparse features

Craigacp:
### Description
This PR adds implementations of `observeSparse` to all the transformers (apart from `IDFTransformation` which already had it). As a consequence to preserve the 4.0 behaviour the transformation fitting methods grow a new argument which turns on or off the use of the `observeSparse` method. It also switches `DoubleFieldProcessor` over so that it always emits values if they are parsable doubles. Before it would elide zero values, but this makes it much harder to implement transformations which should touch every value because they are numerical (rather than categorical encoded as a double).

One downside of this implementation is that it's not possible to change the behaviour of `observeSparse` on a per feature basis (due to it being a tricky breaking API change), which makes it difficult to apply an IDFTransformation to text features while simultaneously transforming features which should ignore the implicit zeros. If this behaviour is required then two `TransformationMap`s can be applied in sequence to a `Dataset`. We'll note this in the release notes for 4.1.

### Motivation
The observeSparse change is so the transformers actually have the documented behaviour, and then that rippled into adding new overloads for the transformation fitting methods. The `DoubleFieldProcessor` change is because it's tricky to correctly transform numerical data without it.

### DIFF
--- a/Core/src/main/java/org/tribuo/transform/TransformTrainer.java
+++ b/Core/src/main/java/org/tribuo/transform/TransformTrainer.java
@@ -51,6 +51,9 @@ public final class TransformTrainer<T extends Output<T>> implements Trainer<T> {
     @Config(description="Densify all the features before applying transformations.")
     private boolean densify;
 
+    @Config(description="Include the implicit zeros in the transformation statistics collection")
+    private boolean observeSparse;
+
     /**
      * For OLCUT.
      */
@@ -60,7 +63,8 @@ public final class TransformTrainer<T extends Output<T>> implements Trainer<T> {
      * Creates a trainer which transforms the data before training, and stores
      * the transformers along with the trained model in a {@link TransformedModel}.
      * <p>
-     * This constructor makes a trainer which keeps the data sparse.
+     * This constructor makes a trainer which keeps the data sparse, and does not use
+     * the implicit zeros to construct the transformations.
      * @param innerTrainer The trainer to use.
      * @param transformations The transformations to apply to the data first.
      */
@@ -71,22 +75,39 @@ public final class TransformTrainer<T extends Output<T>> implements Trainer<T> {
     /**
      * Creates a trainer which transforms the data before training, and stores
      * the transformers along with the trained model in a {@link TransformedModel}.
+     * <p>
+     * Sets {@code observeSparse} to false.
      *
      * @param innerTrainer The trainer to use.
      * @param transformations The transformations to apply to the data first.
      * @param densify Densify the dataset (and any predict time data) before training/prediction.
      */
     public TransformTrainer(Trainer<T> innerTrainer, TransformationMap transformations, boolean densify) {
+        this(innerTrainer,transformations,false,false);
+    }
+
+    /**
+     * Creates a trainer which transforms the data before training, and stores
+     * the transformers along with the trained model in a {@link TransformedModel}.
+     *
+     * @param innerTrainer The trainer to use.
+     * @param transformations The transformations to apply to the data first.
+     * @param densify Densify the dataset (and any predict time data) before training/prediction.
+     * @param observeSparse Use the implicit zeros to construct the transformations.
+     */
+    public TransformTrainer(Trainer<T> innerTrainer, TransformationMap transformations, boolean densify, boolean observeSparse) {
         this.innerTrainer = innerTrainer;
         this.transformations = transformations;
         this.densify = densify;
+        this.observeSparse = observeSparse;
     }
 
     @Override
     public TransformedModel<T> train(Dataset<T> examples, Map<String, Provenance> instanceProvenance) {
         
         logger.fine(String.format("Creating transformers"));
-        TransformerMap transformerMap = examples.createTransformers(transformations);
+
+        TransformerMap transformerMap = examples.createTransformers(transformations,observeSparse);
 
         logger.fine("Transforming data set");
         
--- a/Core/src/main/java/org/tribuo/transform/TransformTrainer.java
+++ b/Core/src/main/java/org/tribuo/transform/TransformTrainer.java
@@ -38,6 +38,7 @@ import java.util.logging.Logger;
  * By default transformations only operate on explicit feature values. To include implicit zeros
  * in transformation fitting set {@code includeImplicitZeroFeatures}. To convert implicit
  * zeros to explicit zeros before applying the transformations set {@code densify}.
+ * See {@link org.tribuo.transform} for a more detailed discussion of these parameters.
  */
 public final class TransformTrainer<T extends Output<T>> implements Trainer<T> {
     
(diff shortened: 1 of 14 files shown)


Core/src/main/java/org/tribuo/transform/TransformTrainer.java
```diff
@@ -35,8 +35,9 @@
  * A {@link Trainer} which encapsulates another trainer plus a {@link TransformationMap} object
  * to apply to each {@link Dataset} before training each {@link Model}.
  * <p>
- * Transformations only operate on observed values. To operate on implicit zeros then
- * first call {@link MutableDataset#densify} on the datasets.
+ * By default transformations only operate on explicit feature values. To include implicit zeros
+ * in transformation fitting then set {@code includeImplicitZeroFeatures}, and to convert implicit
+ * zeros to explicit zeros before applying the transformations set {@code densify}.
```
JackSullivan, line 40:
```suggestion
 * in transformation fitting set {@code includeImplicitZeroFeatures}. To convert implicit
 * zeros to explicit zeros before applying the transformations set {@code densify}.
```

Core/src/main/java/org/tribuo/transform/transformations/MeanStdDevTransformation.java
```diff
@@ -160,13 +162,25 @@ public void observeValue(double value) {
...
         @Override
-        public void observeSparse() { }
+        @Deprecated
+        public void observeSparse() {
+            observeValue(0.0);
+        }
 
         @Override
-        public void observeSparse(int count) { }
+        public void observeSparse(int sparseCount) {
```
JackSullivan, line 171: Why is this copied here and in `BinningTransformer`?

JackSullivan (commented):
This has the changes that we discussed on Friday, but looking over it now, the implicit zeroes vs. densify behavior interactions aren't really explicitly characterized anywhere. I think it would be a good idea to have one place that clearly explains how all four possible settings work that can referred to for more detail in all the various constructors/methods where the interaction is important.

Core/src/main/java/org/tribuo/transform/transformations/MeanStdDevTransformation.java
Craigacp: Do you mean I should refactor out the update logic? That would involve introducing some class to carry the multiple returns. The reason it's in both is because binning transformer can convert things into bins and this one standardises the data, so it's still real valued in this case and implicitly categorical in the binning transformer case.

Craigacp:
Ok, sure. I'll add something to the package-info in `org.tribuo.transform`.

Core/src/main/java/org/tribuo/transform/transformations/MeanStdDevTransformation.java
JackSullivan: that's fair enough. I wasn't clear on the purpose of the two implementations, but if one is rescaling and the other is binning then it's probably simple enough (and the logic is simple enough) to copy the logic.

Core/src/main/java/org/tribuo/transform/package-info.java
```diff
@@ -30,5 +30,39 @@
  * TransformationMap and an inner {@link org.tribuo.Trainer} and produces a
  * {@link org.tribuo.transform.TransformedModel} which automatically transforms it's input data at
  * prediction time.
+ * <p>
+ * Transformations don't produces new {@link org.tribuo.Feature}s - they only modify the values of existing ones.
```
JackSullivan, line 34:
```suggestion
 * Transformations don't produce new {@link org.tribuo.Feature}s - they only modify the values of existing ones.
```

Core/src/main/java/org/tribuo/transform/package-info.java
```diff
@@ -30,5 +30,39 @@
...
+ * not existing at all. Independently, we can explicitly add zero-valued Features by densifying the dataset
+ * before the transformation is fit or before it is applied. Once they exist these Features can be altered by
+ * {@link org.tribuo.transform.Transformer}s and are visible to {@link org.tribuo.transform.Transformation}s which are
+ * being fit.
+ * <p>
+ * The transformation fitting methods have two parameters which alter their behaviour: {@code includeImplicitZeroFeatures}
+ * and {@code densify}. {@code includeImplicitZeroFeatures} controls if the transformation incorporates the implicit zero
+ * valued features (i.e., the ones not present in the example but are present in the dataset's
+ * {@link org.tribuo.FeatureMap}) when building the transformation statistics. This is
+ * important when working with {@link org.tribuo.transform.transformations.IDFTransformation} as it allows correct
```
JackSullivan, line 45:
```suggestion
 * important when working with, e.g. {@link org.tribuo.transform.transformations.IDFTransformation} as it allows correct
```

Core/src/main/java/org/tribuo/transform/package-info.java
```diff
@@ -30,5 +30,39 @@
...
+ * computation of the inverse document frequency, but can be detrimental to features which are one-hot encodings of
+ * categoricals (as they have many more implicit zeros). {@code densify} controls if the example or dataset should have
+ * it's implicit zero valued features converted into explicit zero valued features (i.e., it makes a sparse example into
```
JackSullivan, line 48:
```suggestion
 * its implicit zero valued features converted into explicit zero valued features (i.e., it makes a sparse example into
```

Core/src/main/java/org/tribuo/transform/package-info.java
```diff
@@ -30,5 +30,39 @@
...
+ *     the transformation, but not modified when the transformation is applied. This is most useful when working with
+ *     text data where you want to compute IDF style statistics</li>
+ *     <li>{@code includeImplicitZeroFeatures} is false, {@code densify} is true: the implicit zeros are not used to
+ *     fit the transformation, but are converted to explicit zeros and transformed. This is less useful than the other
+ *     three combinations, but could be used to move the minimum value, or when zero is not appropriate for a missing
+ *     value and needs to be transformed.</li>
+ * </ul>
+ * One further option is to call {@link org.tribuo.MutableDataset#densify} before passing the data to
+ * {@link org.tribuo.transform.TransformTrainer#train}, which is equivalent to setting {@code includeImplicitZeroFeatures}
+ * to true and {@code densify} to true.
```
JackSullivan, line 66:
```suggestion
 * to true and {@code densify} to true. To sum up, in the context of transformations {@code includeImplicitZeroFeatures}
 * determines whether (implicit) zero-values features are <i>measured</i> and {@code densify} determines whether
 * they can be <i>altered</i>. 
```

JackSullivan (changes requested):
Some nitpicky word-choice changes, but otherwise it looks good to me.

Craigacp:
Thanks for the updates.

JackSullivan (approved):

PR #129, opened 2021-04-07
Adds a way to create an ensemble out of existing models

Craigacp:
### Description
This PR adds a method to `WeightedEnsembleModel` called `createEnsembleFromExistingModels` which builds an ensemble out of the supplied model list, ensemble combiner and optionally model combination weights. It adds a `TimestampTrainerProvenance` which is used as the `TrainerProvenance` for any ensembles created this way. Finally it adds a couple of tests for `VotingCombiner` and `AveragingCombiner` along with a few tests of the new behaviour.

The ensemble building mechanism validates that there are at least two models, that there are equal numbers of models and combination weights and that the output domains for all the supplied models are the same. We could consider adding a check that the feature domains are the same, or have some overlap, but it's plausible that text models trained using data subsampling have different views of the feature space and so I've left that check for future work.

### Motivation
It's currently hard in Tribuo to ensemble together models of different types (e.g. an ensemble of a linear model, a decision tree and a kernel SVM). This kind of arbitrary ensembling is occasionally very powerful, but difficult to express via Tribuo's trainer mechanism, so it's simpler to allow users to build custom ensembles from models they already have.

### Paper reference
Kuncheva, L. I. (2014). Combining pattern classifiers: methods and algorithms.

### DIFF
--- a/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
+++ b/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
@@ -49,7 +49,7 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     protected final EnsembleCombiner<T> combiner;
 
     /**
-     * Constructs a weighted ensemble model.
+     * Constructs an ensemble model which uses uniform weights.
      * @param name The model name.
      * @param provenance The model provenance.
      * @param featureIDMap The feature domain.
@@ -138,7 +138,8 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     /**
      * Creates an ensemble from existing models. The model outputs are combined using uniform weights.
      * <p>
-     * Uses the feature and output domain from the first model.
+     * Uses the feature and output domain from the first model as the ensemble model's domains.
+     * The individual ensemble members use the domains that they contain.
      * <p>
      * If the output domains don't cover the same dimensions then it throws {@link IllegalArgumentException}.
      * @param name The ensemble name.
@@ -154,7 +155,8 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     /**
      * Creates an ensemble from existing models.
      * <p>
-     * Uses the feature and output domain from the first model.
+     * Uses the feature and output domain from the first model as the ensemble model's domains.
+     * The individual ensemble members use the domains that they contain.
      * <p>
      * If the output domains don't cover the same dimensions then it throws {@link IllegalArgumentException}.
      * If the weights aren't the same length as the models it throws {@link IllegalArgumentException}.
--- a/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
+++ b/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
@@ -152,7 +152,7 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     }
 
     /**
-     * Creates an ensemble from existing models. The model outputs are combined using uniform weights.
+     * Creates an ensemble from existing models.
      * <p>
      * Uses the feature and output domain from the first model.
      * <p>
--- a/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
+++ b/Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
@@ -49,6 +49,9 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     protected final EnsembleCombiner<T> combiner;
 
     /**
+     * Unless you are implementing a {@link org.tribuo.Trainer} you should
+     * not use this constructor directly. Instead use {@link #createEnsembleFromExistingModels(String, List<Model<T>>, EnsembleCombiner<T>)}.
+     * <p>
      * Constructs an ensemble model which uses uniform weights.
      * @param name The model name.
      * @param provenance The model provenance.
@@ -64,7 +67,10 @@ public final class WeightedEnsembleModel<T extends Output<T>> extends EnsembleMo
     }
 
     /**
-     * Constructs a weighted ensemble model.
+     * Unless you are implementing a {@link org.tribuo.Trainer} you should
+     * not use this constructor directly. Instead use {@link #createEnsembleFromExistingModels(String, List<Model<T>>, EnsembleCombiner<T>, float[])}.
+     * <p>
+     * Constructs an ensemble model which uses uniform weights.
      * @param name The model name.
      * @param provenance The model provenance.
      * @param featureIDMap The feature domain.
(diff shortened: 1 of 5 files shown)


Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -112,4 +134,78 @@ public WeightedEnsembleModel(String name, EnsembleModelProvenance description, I
...
+     * @param combiner The combination function.
+     * @param <T> The output type.
+     * @return A weighted ensemble model.
+     */
+    public static <T extends Output<T>> WeightedEnsembleModel<T> createEnsembleFromExistingModels(String name, List<Model<T>> models, EnsembleCombiner<T> combiner) {
+        return createEnsembleFromExistingModels(name,models,combiner,Util.generateUniformVector(models.size(), 1.0f/models.size()));
+    }
+
+    /**
+     * Creates an ensemble from existing models. The model outputs are combined using uniform weights.
```
JackSullivan, line 155:
```suggestion
     * Creates an ensemble from existing models.
```

Presumably this is just copy-pasted from the above docs.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -112,4 +134,78 @@ public WeightedEnsembleModel(String name, EnsembleModelProvenance description, I
...
+     * <p>
+     * Uses the feature and output domain from the first model.
```
JackSullivan, line 157: 'Uses the feature domain from the first model' could be read to mean that it excludes features not found in that model when creating the ensemble. Is that true? If not I prefer

```suggestion
     * Uses the output domain from the first model. Though the ensemble itself uses the union 
     * of feature domains from the input models, only the first model's feature domain is recorded
     * in provenance.
```

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -45,16 +48,35 @@
 
     protected final EnsembleCombiner<T> combiner;
 
-    public WeightedEnsembleModel(String name, EnsembleModelProvenance description, ImmutableFeatureMap featureIDMap,
+    /**
+     * Constructs a weighted ensemble model.
```
JackSullivan, line 52: In what sense is the 'weighted' if the weights are uniform, or am i missing something?

JackSullivan (commented):
Mostly looks good, apart from a few questions/edits.

I'd like to entertain doing something better with the provenance than just taking the first model's feature domain. i.e., how involved would it be to create a `DatasetUnionProvenance`? We don't seem to need the actual union here, though.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: Oh yeah, this one isn't. I think I copy/pasted it from the one below and changed the relevant params. It's a little weird as `EnsembleModel` is abstract, and so in practice if you want an unweighted combination you need to use `WeightedEnsembleModel` and this constructor. It should have a better name, but renaming things breaks compatibility.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: Yep, thanks.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: I was leaving a bit of wiggle room here. There's an optimization I'm considering where we add `Prediction<T> predict(SGDVector vector)` to the models and then only pay the vector construction cost (and it's associated hashmap lookups and other things) once. I agree that that might make models created here invalid, and I guess I could lock it down so it throws if the feature domains are different?

Craigacp:
Well the ensemble provenance contains each of the individual models' provenances, and those each contain their training dataset provenance. So the information is all there, but I needed to pick something to put in the top level model provenance and the first one seemed easiest.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: I've updated the text.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: Fixed.

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -45,16 +48,35 @@
 
     protected final EnsembleCombiner<T> combiner;
 
-    public WeightedEnsembleModel(String name, EnsembleModelProvenance description, ImmutableFeatureMap featureIDMap,
+    /**
+     * Constructs an ensemble model which uses uniform weights.
```
JackSullivan, line 52: Since user's aren't generally going to be getting `WeightedEnsembleModel`s through this constructor, we should state that.

```suggestion
     * Unless you are implementing a {@link org.tribuo.Trainer} you should
     * not use this constructor directly. Instead use {@link #createEnsembleFromExistingModels(String, List<Model<T>>, EnsembleCombiner<T>)}.
     * <p>
     * Constructs an ensemble model which uses uniform weights.
```

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -45,16 +48,35 @@
...
+    public WeightedEnsembleModel(String name, EnsembleModelProvenance provenance, ImmutableFeatureMap featureIDMap,
                                  ImmutableOutputInfo<T> outputIDInfo,
                                  List<Model<T>> newModels, EnsembleCombiner<T> combiner) {
-        this(name,description,featureIDMap,outputIDInfo,newModels, combiner, Util.generateUniformVector(newModels.size(), 1.0f/newModels.size()));
+        this(name,provenance,featureIDMap,outputIDInfo,newModels, combiner, Util.generateUniformVector(newModels.size(), 1.0f/newModels.size()));
     }
 
-    public WeightedEnsembleModel(String name, EnsembleModelProvenance description, ImmutableFeatureMap featureIDMap,
+    /**
+     * Constructs a weighted ensemble model.
```
JackSullivan, line 67: As above


```suggestion
     * Unless you are implementing a {@link org.tribuo.Trainer} you should
     * not use this constructor directly. Instead use {@link #createEnsembleFromExistingModels(String, List<Model<T>>, EnsembleCombiner<T>, float[])}.
     * <p>
     * Constructs an ensemble model which uses uniform weights.
```

JackSullivan (commented):
Looks good based on our discussion with the above changes.

JackSullivan (approved):
Looks good to me.

PR #132, opened 2021-04-15
Bump OpenCSV to 5.4

### LINKED ISSUE #131: use maven enforcer to ensure dependency convergence
If you run the following command from the top-level directory of Tribuo you get errors:

`mvn enforcer:enforce -Drules=dependencyConvergence`

The first dependency convergence error reported is:

```
+-org.tribuo:tribuo-data:4.1.0-SNAPSHOT
  +-com.opencsv:opencsv:5.2
    +-org.apache.commons:commons-lang3:3.10

```
and
```
+-org.tribuo:tribuo-data:4.1.0-SNAPSHOT
  +-com.opencsv:opencsv:5.2
    +-org.apache.commons:commons-text:1.8
      +-org.apache.commons:commons-lang3:3.9
```

We should probably specify the maven enforcer plugin in the pom.xml file with a snippet like this:

```
 <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
         <version>${maven-enforcer-plugin.version}</version>
         <executions>
           <execution>
             <id>enforce</id>
             <configuration>
               <rules>
                 <dependencyConvergence/>
               </rules>
             </configuration>
             <goals>
               <goal>enforce</goal>
             </goals>
           </execution>
         </executions>
       </plugin>
```

See https://maven.apache.org/enforcer/maven-enforcer-plugin/usage.html for more details and options.

Craigacp:
### Description
OpenCSV 5.2 has a dependency conflict with itself (it pulls in two versions of Apache Commons Lang 3). This bumps the dependency to 5.4 which doesn't have the issue, and also adds the Maven Enforcer plugin to fix #131.

### Motivation
Depending on multiple versions of the same library is likely to cause trouble when deploying Tribuo, even if it's transitively.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,6 +243,19 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-enforcer-plugin</artifactId>
                 <version>3.0.0-M3</version>
+                <executions>
+                    <execution>
+                        <id>enforce</id>
+                        <configuration>
+                            <rules>
+                                <dependencyConvergence/>
+                            </rules>
+                        </configuration>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>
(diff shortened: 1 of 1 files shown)


pogren (approved):
This looks very straightforward and poses little risk if the tests pass.  I have tested the same xml in my local project and this should work fine.

PR #133, opened 2021-04-16
Upgrades OLCUT to 5.1.6

Craigacp:
### Description
Bumps OLCUT to 5.1.6 and migrates ModelProvenance.maybeExtractProvenance over to ObjectProvenance.maybeExtractProvenance.

### Motivation
Pulls in updated jline3 and jackson dependencies and a few OLCUT bug fixes.


pogren (commented):
This all looks straightforward.  Removing the public method from ModelProvenance is fine because it was never in a release.

pogren (approved):
This all looks straightforward.  Removing the public method from ModelProvenance is fine because it was never in a release.

PR #134, opened 2021-04-20
Refactoring AbstractLinearSGDModel and Trainer to extract SGD base classes

Craigacp:
### Description
This PR extracts out `AbstractSGDModel` and `AbstractSGDTrainer` from `AbstractLinearSGDModel` and `AbstractLinearSGDTrainer`, and introduces a `FeedForwardParameters` which has predict and gradient methods. They don't land on `Parameters` because that's used in the CRF as well, and sequences have a differently shaped input.

`AbstractSGDTrainer` has a lot of generic parameters, but those are all hidden from users and the concrete subclasses are still typed with just the output type like most of Tribuo. The code is tested by the existing tests, and can still deserialize Tribuo 4.0 models.

### Motivation
The recent introduction of AbstractLinearSGDModel/Trainer wasn't quite abstract enough. The SGD package could be used for things beyond linear models like factorization machines. This PR will make it straightforward to subclass `AbstractSGDTrainer` for a different model class so you don't have to reimplement the training loop.

### DIFF
--- a/Classification/SGD/src/main/java/org/tribuo/classification/sgd/linear/LinearSGDModel.java
+++ b/Classification/SGD/src/main/java/org/tribuo/classification/sgd/linear/LinearSGDModel.java
@@ -47,7 +47,7 @@ public class LinearSGDModel extends AbstractLinearSGDModel<Label> {
 
     private final VectorNormalizer normalizer;
 
-    // Unused as the weights now live in AbstractLinearSGDModel
+    // Unused as the weights now live in AbstractSGDModel
     // It remains for serialization compatibility with Tribuo 4.0
     @Deprecated
     private DenseMatrix weights = null;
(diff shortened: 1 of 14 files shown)


pogren (approved):
Introduces AbstractSGDModel which basically moves AbstractLinearSGDModel methods 'predictSingle' and the supporting inner class PredAndActive into this class.  Also introduces AbstractSGDTrainer which steals most of its code from AbstractLinearSGDTrainer which is now basically delegates to its super class.  Two protected constructors were removed from AbstractLinearSGDTrainer so this is not a backwards compatible change and will affect subclasses of AbstractLinearSGDTrainer (if they exist.)  

LinearParameters now implements FeedForwardParameters which defines a 'predict' method whose return type is DenseVector which LinearParameters now implements.  Previously LinearParameters implemented Parameters directly and defined its own 'predict' method which returned an SGDVector - so this is not a backwards compatible change and will affect subclasses of LinearParameters (if they exist.)  The methods 'predict' and 'gradients' are now annotated with '@Override' because they are defined in FeedForwardParameters which also introduces the 'copy' method.  

Other than the above noted concerns - this is a straightforward refactoring of the abstract super classes of the SGD model and training code to better share code.

PR #136, opened 2021-04-27
feature/pretty-MultiLabelConfusionMatrix-toString

nezda:
### Description
Easier to read `MultiLabelConfusionMatrix.toString()`. (Subset of PR [#128] which includes other work.)

### Motivation

Such a feature would make analysis of `MultiLabel` models simpler: in particular `MultiLabelConfusionMatrix` `toString()` is hard to interpret. Compare

```
	row 0 [ 3.000000000000000, 0.000000000000000];
	row 1 [ 0.000000000000000, 1.000000000000000];
)
	row 0 [ 0.000000000000000, 1.000000000000000];
	row 1 [ 1.000000000000000, 2.000000000000000];
)
	row 0 [ 2.000000000000000, 1.000000000000000];
	row 1 [ 1.000000000000000, 0.000000000000000];
)
]
```

to

```
(LabelSet={a})
    [tn: 3 fn: 0]
    [fp: 0 tp: 1]
(LabelSet={b})
    [tn: 0 fn: 1]
    [fp: 1 tp: 2]
(LabelSet={c})
    [tn: 2 fn: 1]
    [fp: 1 tp: 0]
```

cc @Craigacp

### DIFF
--- a/MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
+++ b/MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
@@ -28,6 +28,7 @@ import org.tribuo.multilabel.MultiLabelFactory;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * A {@link ConfusionMatrix} which accepts {@link MultiLabel}s.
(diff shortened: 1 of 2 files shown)


MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
```diff
@@ -158,15 +159,18 @@ public double confusion(MultiLabel predicted, MultiLabel truth) {
...
-        StringBuilder sb = new StringBuilder();
-        sb.append("[");
-        for (int i = 0; i < mcm.length; i++) {
-            DenseMatrix cm = mcm[i];
-            sb.append(cm.toString());
-            sb.append("\n");
-        }
-        sb.append("]");
-        return sb.toString();
+        return getDomain().getDomain().stream()
```
Craigacp, line 162: This is sensitive to the iteration order of the HashSet inside the domain, whereas the old one was fixed on construction (though when it was constructed it was fixed based on the iteration order of the HashSet in the domain).

This comment is more a todo for later when the label domains have consistent iteration order (probably lexicographic like the features).

Craigacp (approved):
LGTM, thanks.

Craigacp:
I'm trying to figure out why the CI hasn't run, once I've sorted that out and it passes I'll merge in this PR.

Craigacp:
Seems like the recent change to GH Actions which limits PR workflow runs from new contributors is causing trouble. I've ran the CI tests locally and everything passes, so I'm going to merge this now.

MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
nezda: Ah, good point. That variable iteration order has confused me several times. There's always `LinkedHashSet`? Or maybe https://docs.oracle.com/javase/9/docs/api/java/util/Set.html#of-E...- would have stable iteration order like Guava's `ImmutableHashSet`? How long will Java 8 be required?

PR #137, opened 2021-04-27
feature/RowProcessor-replaceNewlinesWithSpaces

nezda:
### Description
Add optional `boolean` flag `RowProcessor.replaceNewlinesWithSpaces` that controls whether or not to replace newlines with spaces in values before passing them to field processors (don't change default behavior by defaulting this to `true`).

### Motivation
`TextFieldProcessor`s may  wish to be aware of newlines in the input and encode them as features. Example encoded as a unit test.

### DIFF
--- a/Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
@@ -86,7 +86,7 @@ public class RowProcessor<T extends Output<T>> implements Configurable, Provenan
     protected Map<String,FieldProcessor> regexMappingProcessors = new HashMap<>();
 
     @Config(description="Replace newlines with spaces in values before passing them to field processors.")
-    protected boolean replaceNewlinesWithSpaces;
+    protected boolean replaceNewlinesWithSpaces = true;
 
     protected boolean configured;
 
@@ -385,7 +385,7 @@ public class RowProcessor<T extends Output<T>> implements Configurable, Provenan
             String value = row.get(e.getKey());
             if (value != null) {
                 if (replaceNewlinesWithSpaces) {
-                  value = value.replace('\n', ' ');
+                    value = value.replace('\n', ' ');
                 }
                 value = value.trim();
                 features.addAll(e.getValue().process(value));
@@ -562,7 +562,7 @@ public class RowProcessor<T extends Output<T>> implements Configurable, Provenan
      */
     @Deprecated
     public RowProcessor<T> copy() {
-      return new RowProcessor<>(metadataExtractors, weightExtractor, responseProcessor, fieldProcessorMap, regexMappingProcessors, featureProcessors, replaceNewlinesWithSpaces);
+        return new RowProcessor<>(metadataExtractors, weightExtractor, responseProcessor, fieldProcessorMap, regexMappingProcessors, featureProcessors, replaceNewlinesWithSpaces);
     }
 
     @Override
--- a/Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
@@ -85,6 +85,9 @@ public class RowProcessor<T extends Output<T>> implements Configurable, Provenan
     @Config(description="A map from a regex to field processors to apply to fields matching the regex.")
     protected Map<String,FieldProcessor> regexMappingProcessors = new HashMap<>();
 
+    @Config(description="Replace newlines with spaces in values before passing them to field processors.")
+    protected boolean replaceNewlinesWithSpaces;
+
     protected boolean configured;
 
     /**
@@ -154,7 +157,7 @@ public class RowProcessor<T extends Output<T>> implements Configurable, Provenan
     public RowProcessor(List<FieldExtractor<?>> metadataExtractors, FieldExtractor<Float> weightExtractor,
                         ResponseProcessor<T> responseProcessor, Map<String,FieldProcessor> fieldProcessorMap,
                         Set<FeatureProcessor> featureProcessors) {
-        this(metadataExtractors,weightExtractor,responseProcessor,fieldProcessorMap,Collections.emptyMap(),featureProcessors);
+        this(metadataExtractors,weightExtractor,responseProcessor,fieldProcessorMap,Collections.emptyMap(),featureProcessors, true);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
```diff
@@ -85,6 +85,9 @@
     @Config(description="A map from a regex to field processors to apply to fields matching the regex.")
     protected Map<String,FieldProcessor> regexMappingProcessors = new HashMap<>();
 
+    @Config(description="Replace newlines with spaces in values before passing them to field processors.")
+    protected boolean replaceNewlinesWithSpaces;
```
Craigacp, line 89: This needs to be set to `true` otherwise old configs will load with it set to `false` and break config compatibility.

Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
```diff
@@ -349,7 +384,10 @@ public void postConfig() {
         for (Map.Entry<String,FieldProcessor> e : fieldProcessorMap.entrySet()) {
             String value = row.get(e.getKey());
             if (value != null) {
-                value = value.replace('\n', ' ').trim();
+                if (replaceNewlinesWithSpaces) {
+                  value = value.replace('\n', ' ');
```
Craigacp, line 388: This line has 2 spaces rather than 4.

Data/src/main/java/org/tribuo/data/columnar/RowProcessor.java
```diff
@@ -524,7 +562,7 @@ public void expandRegexMapping(Collection<String> fieldNames) {
      */
     @Deprecated
     public RowProcessor<T> copy() {
-        return new RowProcessor<>(metadataExtractors, weightExtractor, responseProcessor, fieldProcessorMap, regexMappingProcessors, featureProcessors);
+      return new RowProcessor<>(metadataExtractors, weightExtractor, responseProcessor, fieldProcessorMap, regexMappingProcessors, featureProcessors, replaceNewlinesWithSpaces);
```
Craigacp, line 565: This line has two spaces rather than 4.

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
```diff
@@ -170,4 +183,127 @@ public void metadataExtractorTest() {
...
 
+    @Test
+    public void replaceNewlinesWithSpacesTest() {
+        final Pattern BLANK_LINES = Pattern.compile("(\n[\\s-]*\n)+");
+
+        final Function<CharSequence, CharSequence> newLiner = (CharSequence charSequence) -> {
+            if (charSequence == null || charSequence.length() == 0) {
+                return charSequence;
+            }
+            return BLANK_LINES.splitAsStream(charSequence).collect(Collectors.joining(" ⏎\n\n"));
```
Craigacp, line 194: What's the character before the `\n\n` here?

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
```diff
@@ -170,4 +183,127 @@ public void metadataExtractorTest() {
...
+        a = featureIterator.next();
+        assertEquals("order_text@1-N=Jimmy", a.getName());
+        a = featureIterator.next();
+        assertEquals("order_text@1-N=⏎", a.getName());
+        a = featureIterator.next();
+        assertEquals("order_text@2-N=Jimmy/⏎", a.getName());
+        a = featureIterator.next();
+        assertEquals("order_text@2-N=⏎/Hoffa", a.getName());
+        assertFalse(featureIterator.hasNext());
+        a = featureIterator.next(); // weird this doesn't throw NoSuchElementException?
```
Craigacp, line 229: Yep, that's a bug in the iterator on `ArrayExample`, it throws `ArrayIndexOutOfBoundsException` when exhausted rather than `NoSuchElementException`. We'll fix it.

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
```diff
@@ -170,4 +183,127 @@ public void metadataExtractorTest() {
...
+        assertNull(a.getName());
+    }
+
+    public abstract static class ForwardingTokenizer implements Tokenizer {
```
Craigacp, line 233: Could this be folded into `MungingTokenizer`? It doesn't seem to be used anywhere else.

Craigacp (changes requested):

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
Craigacp: Fixed in #138.

Craigacp: Ah, actually you were hitting a slightly different issue here, which is that the backing array for an `ArrayExample` can be larger than the size of that example, and the iterator was happy to read past size if you kept calling `next` without `hasNext` because the check wasn't correct. When #138 is integrated this line will correctly throw `NoSuchElementException`.

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
nezda: https://graphemica.com/%E2%8F%8E kinda picked it for fun

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
nezda: sure, i just copy+pasted in from my project

nezda: i wonder if those might be useful to others?

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
Craigacp: Ok. Given this is about newline splitting it would probably be better if it wasn't a return symbol in case there's some font glyph replacement on the newlines. So any other sensible emoji is fine, but the return symbol might be confusing.

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
Craigacp: Well we can consider expanding the `Tokenizer` implementations in a different PR, but for the purposes of the test lets minimise the code that's added.

nezda: 👍  done

Data/src/test/java/org/tribuo/data/columnar/RowProcessorTest.java
```diff
@@ -170,4 +183,114 @@ public void metadataExtractorTest() {
...
+        final Map<String, FieldProcessor> fieldProcessors = new HashMap<>();
+        fieldProcessors.put("order_text", new TextFieldProcessor("order_text", textPipeline));
+
+        MockResponseProcessor response = new MockResponseProcessor("Label");
+
+        Map<String,String> row = new HashMap<>();
+        row.put("order_text", "Jimmy\n\n\n\nHoffa");
+        row.put("Label", "Sheep");
+
+        RowProcessor<MockOutput> processor = new RowProcessor<>(Collections.emptyList(),null,response,fieldProcessors,Collections.emptyMap(),Collections.emptySet(), false);
```
Craigacp, line 209: I think it would be a good idea to have this also test with the flag turned on on the same example so it verifies that the features emitted are different.

nezda: done

JackSullivan (approved):
This looks good to me. I originally wrote this, and revisiting it now, I'm not convinced that the default behavior is correct. That is, we should not be altering the incoming data in `RowProcessor` before it is fed to the feature processors. In a future version we'll probably want to change that, but to avoid breaking changes this is probably the best interim solution. The constructor situation on `RowProcessor` is also in great need of harmonization, probably through consolidation into a builder pattern.

Thanks for putting in the work to put it in, and for bringing this code to our attention again.

Craigacp (approved):

Craigacp:
Thanks. As Jack said, we probably want to revisit this as part of a larger refactor to make `RowProcessor` a little simpler to use, but in the meantime this seems like a fine change.

PR #138, opened 2021-04-27
Fixing ArrayExample.iterator so it throws NoSuchElementException

Craigacp:
### Description
ArrayExample.iterator and BinaryFeaturesExample.iterator used to throw ArrayIndexOutOfBoundsException rather than the expected NoSuchElementException when exhausted. Now they throw the correct exception.

### Motivation
The iterators should conform to the documentation for `java.util.Iterator`.

### DIFF
--- a/Core/src/main/java/org/tribuo/impl/ArrayExample.java
+++ b/Core/src/main/java/org/tribuo/impl/ArrayExample.java
@@ -34,6 +34,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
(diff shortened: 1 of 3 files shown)


pogren (approved):
this is a very straightforward update that looks good.  iterator is now correctly throwing a NoSuchElementException rather than just sort of accidentally throwing an index out of bounds exception....

Craigacp:
This actually also fixes a less obvious bug, which is that ArrayExample.iterator was happy to read up until the end of the backing array if you kept calling `next`, even though the backing array could be larger than the number of features. This would cause it to return feature objects with junk in them. The test also catches this case.

PR #139, opened 2021-05-11
Small fixes mostly for javadoc

### LINKED ISSUE #108: How do we unpack a DimensionTuple?
**Ask the question**
I'm trying to unpack the prediction results after calling `model.predict(data)`, I get a `ArrayList` of `Prediction` items.

Each `Prediction` item in my case is of type `Regressor.DimensionTuple`, there is a `getOutput()` method  and a `getValues()` on it but neither give me the prediction value (numeric). It seems there isn't an easy method/API to get this out.

When I call this `xgbModel.predict(data).get(0).getOutput()`, I get this `(DIM-0,5.982707500457764)`
When I call this `xgbModel.predict(data).get(0).getOutput().getValues()`, I get this `[D@6b788c95`, the `getDimensionNamesString()` works but not the one to get the value corresponding to it. 

What am I not doing? This I believe is an important function. I have been looking at the docs at  https://tribuo.org/learn/4.0/javadoc/org/tribuo/regression/Regressor.DimensionTuple.html

**Is your question about a specific Tribuo class?**
Class Regressor.DimensionTuple, 

**Additional context**
Using the prediciton results to do further analysis.

Craigacp:
### Description
- Added more javadoc, covering most public methods and classes. 
- Expanded the docs in `Regressor` to make it more clear how it behaves.
- Added a new method to `Regressor` to get the `DimensionTuple` by index.
- Fixed typos and malformed javadoc in several places.
- Fixed `Prediction.toString` so it doesn't truncate the last few characters.
- Added `toString` to `AnomalyInfo`, `ClusteringInfo` and `MutableLabelInfo` which were previously missing.
- Added some missing constructors to the tree packages to preserve backwards compatibility with 4.0.
- Added extra trees to ClassificationEnsembleOptions.
- Renamed a few of the options inner classes inside `TrainTest` as their names were incorrect. This is technically a breaking change, but subclassing options is not the idiomatic way of using them, and they are inner classes to denote that they are for this specific instance of `TrainTest`.

### Motivation
Tribuo's documentation is a little sparse, especially around the behaviour of regressors which is a little different to most packages.

The Regressor doc updates fix #108.


pogren (approved):
The changes in this PR almost exclusively related to javadoc and related formatting changes.  The only change that looks remotely like a bug fix is WeightedEnsembleModel.createEnsembleFromExistingModels whose return value now makes use of the 'name' parameter.  Other minor changes include refactoring toString methods, adding a getter called getDimension(int) to Regressor, throwing an UncheckedIOException instead of a RuntimeException in Wordpiece, adding constructors (e.g. CARTRegressionTrainer), and renaming the options definitions in some of the TrainTest drivers such as e.g. LibLinearOptions to LibSVMOptions in libsvm.TrainTest.java.

PR #141, opened 2021-05-24
Adding the RNG to AbstractSGDTrainer.createParameters()

Craigacp:
### Description
The parameter creation hook in `AbstractSGDTrainer` now accepts a RNG to allow for random initialisations of the logistic regression and future trainers which depend on `AbstractSGDTrainer`.

### Motivation
Tribuo's linear sgd models use a fixed initialisation, but we're considering making it random. Some of the other SGD models we're investigating for the next release will need random initialisations for symmetry breaking, and so they need access to the localRNG produced for that call to train. The RNG state is already tracked in the trainer provenance so this change is very small.


JackSullivan (approved):
Looks good to me.

PR #148, opened 2021-06-30
Adds a field processor which operates on dates

Craigacp:
### Description
Adds a `FieldProcessor` which parses a date value into a variety of features.

This also adds a new `GeneratedFeatureType` as date values produce integer features, along with adding locale generation to `DateExtractor` and `OffsetDateTimeExtractor`.

### Motivation
Tribuo needs a way of using dates as features as well as example metadata.

### DIFF
--- a/Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
@@ -23,6 +23,7 @@ import com.oracle.labs.mlrg.olcut.provenance.impl.ConfiguredObjectProvenanceImpl
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -39,6 +40,12 @@ public class DateExtractor extends SimpleFieldExtractor<LocalDate> {
     private String dateFormat;
     private DateTimeFormatter formatter;
 
+    @Config(mandatory = false, description = "Sets the locale language.")
+    private String localeLanguage = null;
+
+    @Config(mandatory = false, description = "Sets the locale country.")
+    private String localeCountry = null;
+
     /**
      * for olcut
      */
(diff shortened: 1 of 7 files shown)


Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
```diff
@@ -77,8 +102,14 @@ public DateExtractor(String fieldName, String metadataName, DateTimeFormatter fo
     @Override
     public void postConfig() {
         super.postConfig();
+        Locale locale;
+        if ((localeLanguage == null) && (localeCountry == null)) {
+            locale = Locale.getDefault(Locale.Category.FORMAT);
+        } else {
+            locale = new Locale(localeLanguage,localeCountry);
```
JackSullivan, line 109: Won't we end up here if one of `localeLanguage` and `localeCountry` are null?
Maybe if either one is null we use the default and log a warning.

Data/src/main/java/org/tribuo/data/columnar/extractors/OffsetDateTimeExtractor.java
```diff
@@ -65,8 +88,14 @@ public OffsetDateTimeExtractor(String fieldName, String metadataName, String dat
     @Override
     public void postConfig() {
         super.postConfig();
+        Locale locale;
+        if ((localeLanguage == null) && (localeCountry == null)) {
+            locale = Locale.getDefault(Locale.Category.FORMAT);
+        } else {
+            locale = new Locale(localeLanguage,localeCountry);
```
JackSullivan, line 95: Won't we end up here if one of `localeLanguage` and `localeCountry` are null?
Maybe if either one is null we use the default and log a warning.

JackSullivan (commented):
Looks good apart from the minor comments inline on Locale nulls.

Data/src/main/java/org/tribuo/data/columnar/extractors/OffsetDateTimeExtractor.java
Craigacp: I changed it so they throw `PropertyException`. The lack of default in locale is to ensure the default behaviour is the same in Tribuo 4.2+ as it was in 4.1 and 4.0.

Data/src/main/java/org/tribuo/data/columnar/extractors/DateExtractor.java
Craigacp: Changed to throw `PropertyException`.

JackSullivan (approved):
Looks good to me

PR #150, opened 2021-07-12
Added support for multioutputs to ResponseProcesser, with tests.

JackSullivan:
### Description
Changes `ResponseProcessor`s to process response inputs for multiple fields, supporting things like multilabel or multiclass classification. 

### Motivation
This is a new feature that extends the generality of the `org.tribuo.data.columnar` system, but the direct motivation is to allow `CSVLoader` to become a special case of `CSVDataSource`, by supporting all of the features of `CSVLoader` within `ColumnarDataSource`.


Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -47,6 +52,36 @@
...
+    private List<String> positiveResponses;
+
+    @Config(description = "Whether to display field names as part of the generated label, defaults to false")
+    private boolean displayField = false;
+
+
+    @Override
+    public void postConfig() {
+        if (fieldName != null && fieldNames != null) { // we can only have one path
+            throw new PropertyException("fieldName, FieldNames", "only one of fieldName or fieldNames can be populated");
```
Craigacp, line 68: The first argument to the three string constructor for `PropertyException` is supposed to be the name of the object in the config file. In practice I don't bother putting in the `ConfigurableName` annotation just to get that information, though if I'd thought more clearly about it I guess it could have been an argument to `postConfig`. Either way, you should probably set it to the empty string to mirror the rest of the places in Tribuo where we don't have this information and throw `PropertyException`.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -47,6 +52,36 @@
...
+        } else if (fieldNames != null) {
+            positiveResponses = positiveResponses == null ? Collections.nCopies(fieldNames.size(), positiveResponse) : positiveResponses;
```
Craigacp, line 70: This should check that `positiveResponse` is non-null before copying it.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -47,6 +52,36 @@
...
+            if(positiveResponses.size() != fieldNames.size()) {
+                throw new PropertyException("positiveResponses", "must either be empty or match the length of fieldNames");
+            }
+        } else if (fieldName != null)  {
+            if(positiveResponses != null) {
+                throw new PropertyException("positiveResponses", "if fieldName is populated, positiveResponses must be blank");
+            }
+            fieldNames = Collections.singletonList(fieldName);
+            positiveResponses = Collections.singletonList(positiveName);
```
Craigacp, line 79: This is `positiveName` not `positiveResponse`? Also it should check that it's not null.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -60,9 +95,66 @@ private BinaryResponseProcessor() {}
...
+     * @param positiveResponse The positive response to look for.
+     * @param outputFactory The output factory to use.
+     */
+    public BinaryResponseProcessor(List<String> fieldNames, String positiveResponse, OutputFactory<T> outputFactory) {
+        this(fieldNames, Collections.nCopies(fieldNames.size(), positiveResponse), outputFactory);
+    }
+
+    /**
+     * Constructs a binary response processor which emits a positive value for a single string
+     * and a negative value for all other field values. the lengths of fieldNames and positiveResponses
```
Craigacp, line 114: s/the/The/

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -60,9 +95,66 @@ private BinaryResponseProcessor() {}
...
+     * @param positiveResponses The positive responses to look for.
+     * @param outputFactory The output factory to use.
+     */
+    public BinaryResponseProcessor(List<String> fieldNames, List<String> positiveResponses, OutputFactory<T> outputFactory) {
+        this(fieldNames, positiveResponses, outputFactory, false);
+    }
+
+    /**
+     * Constructs a binary response processor which emits a positive value for a single string
+     * and a negative value for all other field values. the lengths of fieldNames and positiveResponses
```
Craigacp, line 126: s/the/The/

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -60,9 +95,66 @@ private BinaryResponseProcessor() {}
...
+     * must be the same.
+     * @param fieldNames The field names to read.
+     * @param positiveResponses The positive responses to look for.
+     * @param outputFactory The output factory to use.
+     * @param displayField whether to include field names in the generated labels.
+     */
+    public BinaryResponseProcessor(List<String> fieldNames, List<String> positiveResponses, OutputFactory<T> outputFactory, boolean displayField) {
+        this(fieldNames, positiveResponses, outputFactory, "1", "0", displayField);
```
Craigacp, line 134: These new constructors should probably document what they default `positiveName` and `negativeName` to, which probably implies that "1" and "0" should be public static final constants so you can easily refer to them from the Javadoc.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -60,9 +95,66 @@ private BinaryResponseProcessor() {}
...
+    }
+
+    /**
+     * Constructs a binary response processor which emits a positive value for a single string
+     * and a negative value for all other field values. the lengths of fieldNames and positiveResponses
```
Craigacp, line 139: s/the/The/

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -72,7 +164,7 @@ public BinaryResponseProcessor(String fieldName, String positiveResponse, Output
 
     @Override
     public String getFieldName() {
```
Craigacp, line 166: Should we tag this method deprecated too? I think that means that Maven doesn't warn you for using a deprecated method.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -83,12 +175,30 @@ public void setFieldName(String fieldName) {
 
     @Override
     public Optional<T> process(String value) {
-        return Optional.of(outputFactory.generateOutput(positiveResponse.equals(value) ? positiveName : negativeName));
+        return process(Collections.singletonList(value));
+    }
+
+    @Override
+    public Optional<T> process(List<String> values) {
+        List<String> responses = new ArrayList<>();
```
Craigacp, line 183: Maybe we should check that `values.size()==fieldNames.size()`?

Data/src/main/java/org/tribuo/data/columnar/processors/response/FieldResponseProcessor.java
```diff
@@ -17,28 +17,62 @@
...
+    @Config(description="The field name to read.")
+    @Deprecated
     private String fieldName;
 
-    @Config(mandatory = true,description="Default value to return if one isn't found.")
+    @Config(description="Default value to return if one isn't found.")
     private String defaultValue;
 
-    @Config(mandatory = true,description="The output factory to use.")
+    @Config(description="The output factory to use.")
```
Craigacp, line 44: The output factory should still be mandatory.

Data/src/main/java/org/tribuo/data/columnar/processors/response/FieldResponseProcessor.java
```diff
@@ -17,28 +17,62 @@
...
+
+    @Config(description = "Whether to display field names as part of the generated label, defaults to false")
+    private boolean displayField = false;
+
+    @Override
+    public void postConfig() {
+        if (fieldName != null && fieldNames != null) {
+            throw new PropertyException("fieldName, FieldNames", "only one of fieldName or fieldNames can be populated");
+        } else if (fieldNames != null) {
+            defaultValues = defaultValues == null ? Collections.nCopies(fieldNames.size(), defaultValue) : defaultValues;
```
Craigacp, line 61: Check `defaultValue` is non-null.

Data/src/main/java/org/tribuo/data/columnar/processors/response/FieldResponseProcessor.java
```diff
@@ -17,28 +17,62 @@
...
+            if(defaultValues.size() != fieldNames.size()) {
+                throw new PropertyException("defaultValues", "must either be empty or match the length of fieldNames");
+            }
+        } else if (fieldName != null) {
+            if(defaultValues != null) {
+                throw new PropertyException("defaultValues", "if fieldName is populated, defaultValues must be blank");
+            }
+            fieldNames = Collections.singletonList(fieldName);
+            defaultValues = Collections.singletonList(defaultValue);
```
Craigacp, line 70: Check `defaultValue` is non-null.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -33,9 +39,11 @@
 public class QuartileResponseProcessor<T extends Output<T>> implements ResponseProcessor<T> {
 
     @Config(mandatory = true,description="The string to emit.")
+    @Deprecated
     private String name;
 
     @Config(mandatory = true,description="The field name to read.")
+    @Deprecated
```
Craigacp, line 46: This and `quartile` shouldn't be mandatory anymore.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -44,13 +52,39 @@
...
+
+    @Config(description = "A list of quartiles to use, should have the same length as fieldNames")
+    private List<Quartile> quartiles;
+
+    @Override
+    public void postConfig() throws PropertyException, IOException {
+        if (fieldName != null && fieldNames != null) {
+            throw new PropertyException("fieldName, FieldNames", "only one of fieldName or fieldNames can be populated");
+        } else if (fieldNames != null) {
+            quartiles = quartiles == null ? Collections.nCopies(fieldNames.size(), quartile) : quartiles;
```
Craigacp, line 66: Should check if `quartile` is null.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -44,13 +52,39 @@
...
+            if(quartiles.size() != fieldNames.size()) {
+                throw new PropertyException("quartiles", "must either be empty or match the length of fieldNames");
+            }
+        } else if (fieldName != null) {
+            if (quartiles != null) {
+                throw new PropertyException("quartiles", "if fieldName is populated, quartiles must be blank");
+            }
+            fieldNames = Collections.singletonList(fieldName);
+            quartiles = Collections.singletonList(quartile);
```
Craigacp, line 75: Should check if `quartile` is null.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -33,9 +39,11 @@
 public class QuartileResponseProcessor<T extends Output<T>> implements ResponseProcessor<T> {
 
     @Config(mandatory = true,description="The string to emit.")
+    @Deprecated
```
Craigacp, line 42: Why is `name` deprecated?

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -60,9 +94,22 @@ private QuartileResponseProcessor() {}
...
+     * Constructs a response processor which emits 4 distinct bins for the output factory to process.
+     * <p>
+     * This works best with classification outputs as the discrete binning is tricky to do in other output
+     * types.
+     * @param fieldNames The field to read.
+     * @param quartiles The quartile range to use.
+     * @param outputFactory The output factory to use.
+     */
+    public QuartileResponseProcessor(List<String> fieldNames, List<Quartile> quartiles, OutputFactory<T> outputFactory) {
+        this.fieldNames = fieldNames;
```
Craigacp, line 111: These constructors are a bit weird now. Name used to be mandatory, and it's not clear why it's not anymore.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -101,9 +148,39 @@ public String getFieldName() {
...
+
+    @Override
+    public List<String> getFieldNames() {
+        return fieldNames;
+    }
+
     @Override
     public String toString() {
-        return "QuartileResponseProcessor(fieldName="+ fieldName +",quartile="+quartile.toString()+")";
+        return "QuartileResponseProcessor(fieldNames="+ fieldNames.toString() +",quartiles=" + quartiles.stream().map(Quartile::toString).collect(Collectors.toList()) + ")";
```
Craigacp, line 183: Doesn't `quartiles.toString()` call each quartile's `toString`?

Data/src/test/java/org/tribuo/data/columnar/processors/response/MultioutputResponseProcessorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    private URI missingDataFile;
+    private Map<String, FieldProcessor> fieldProcessors;
+    private List<Set<String>> multiTruthValues;
+
+
+    private static <T extends Output<T>> RowProcessor<T> makeRowProcessor(ResponseProcessor<T> responseProcessor, Map<String, FieldProcessor> fieldProcessors) {
+        return new RowProcessor<T>(responseProcessor, fieldProcessors);
+    }
+
+    private <T extends Output<T>, Label> void doTest(ResponseProcessor<T> responseProcessor, List<Label> expectedLabels, Function<T, Label> labelMapper) {
```
Craigacp, line 42: Could you rename this type parameter from `Label` to something else, preferably all caps. I spent quite a while staring at it trying to figure out how it had access to the `Label` type given we're in Core.

Craigacp (changes requested):
Bunch of small things and a couple of correctness issues.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
JackSullivan: As we discussed offline, the `name` usage (where it appends something to the output) isn't really matched by functionality in any of the other `RowProcessor`s and doesn't align well with multiple output support.

JackSullivan:
I believe I've addressed all the changes requested with this commit, let me know if I missed something.

Data/src/main/java/org/tribuo/data/columnar/ResponseProcessor.java
```diff
@@ -47,13 +50,36 @@
...
      * Returns Optional.empty() if it failed to process out a response.
      * @param value The value to process.
      * @return The response value if found.
      */
+    @Deprecated
     public Optional<T> process(String value);
+
+    /**
+     * Returns Optional.empty() if it failed to process out a response. Though this method currently has
+     * a default implementation, it is the core method and should be overridden by implementors. Unless it
```
Craigacp, line 67: I think I'd prefer "This method has a default implementation for backwards compatibility with Tribuo 4.0 and 4.1. This method should be overridden by code which depends on newer versions of Tribuo. The default implementation will be removed when the deprecated members are removed." to make it explicit.

Data/src/main/java/org/tribuo/data/columnar/processors/response/BinaryResponseProcessor.java
```diff
@@ -32,47 +37,153 @@
...
 
-    @Config(mandatory = true,description="Output factory to use to create the response.")
+    @Config(mandatory = true, description="Output factory to use to create the response.")
     private OutputFactory<T> outputFactory;
 
+    public static final String POSITIVE_NAME = "1";
+    public static final String NEGATIVE_NAME = "0";
+
     @Config(description="The positive response to emit.")
     private String positiveName = "1";
```
Craigacp, line 54: These should default to `POSITIVE_NAME` and `NEGATIVE_NAME`.

Data/src/main/java/org/tribuo/data/columnar/processors/response/QuartileResponseProcessor.java
```diff
@@ -101,9 +166,42 @@ public String getFieldName() {
...
 
+    @Override
+    public Optional<T> process(List<String> values) {
+        if(values.size() != fieldNames.size()) {
+            throw new IllegalArgumentException("values must have the same length as fieldNames. Got values: " + values.size() + " fieldNames: "  + fieldNames.size());
+        }
+        List<String> response = new ArrayList<>();
+        for(int i=0; i< fieldNames.size(); i++) {
+            String value = values.get(i);
+            String prefix = name == null || name.isEmpty() ? fieldNames.get(i) : getFieldName();
```
Craigacp, line 177: Shouldn't the last part of the ternary be `: name`? Otherwise this has different behaviour depending on what process method you call.

Craigacp (changes requested):
I think the quartile one needs another look to make sure the behaviour hasn't changed, but otherwise it's pretty close.

Craigacp (approved):
LGTM.

PR #152, opened 2021-07-15
Adding an ARM maven profile which skips the native library tests

Craigacp:
### Description
XGBoost, ONNX and TF don't provide ARM binaries on Maven Central, so those tests need to be skipped on ARM platforms. This change makes it easier to do so.

It also removes a deprecation warning from `MutableRegressionInfo` which should have switched to using `UNKNOWN_REGRESSOR` when we added the deprecation notice to `UNKNOWN_MULTIPLE_REGRESSOR` in 4.1. Both fields refer to the same object, so there is no visible change.


pogren (approved):
both changes are straightforward and low risk.

PR #153, opened 2021-07-15
CSVLoader refactor

Craigacp:
### Description
This change rebases `CSVLoader` on top of `CSVDataSource`, instead of reading the CSVs itself and processing the examples into a `ListDataSource`. 

There are currently two breaking changes both of which are sufficiently minor that we're ok with introducing them in a minor release. 

- All the `CSVLoader.loadDataSource` overloads now return `DataSource<T>` rather than `ListDataSource<T>` and the source they return is an instance of `CSVDataSource<T>`. This is unavoidable unless we want to read the `CSVDataSource` directly into a `ListDataSource` on construction, which might cause provenance issues.
- `CSVLoader.loadDataSource` is now lazy, like `CSVDataSource` itself. Previously it used to read the whole csv file and buffer the examples in memory, triggering any parsing exceptions during the call. Now the file is read as it's iterated, so the parsing exceptions are triggered on loading into a dataset, and the examples are not cached. We could modify `CSVDataSource` to have a cache flag which recovered the original behaviour.

This work caused a few changes in `DoubleFieldProcessor` and `FieldResponseProcessor` to allow them to behave in the same way as the old `CSVLoader` parsing code. All these options are off by default, so the current behaviour is preserved. I also went through and tagged all the method overrides where the super class method was marked deprecated in PR #150 which removes all the compile time deprecation warnings.

There is one further behaviour change, which is multi-output responses are generated in a different order. This seems to be because the set iteration order is not fixed, and so could have broken the test at any time in the old code. The test was fixed by using a `LinkedHashSet`, and we probably should specify in the docs that this should be used to ensure consistent iteration order.

### Motivation
CSVLoader currently emits `ListDataSource` with a special provenance class. This isn't configurable, and so that means it needs to be special cased in the reproducibility system, which seems unnecessary. After this change the CSVLoader will emit a CSVDataSourceProvenance which can be used to reconstruct the data without user intervention or special casing.

### DIFF
--- a/Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
@@ -67,7 +67,7 @@ public class DoubleFieldProcessor implements FieldProcessor {
      * Constructs a field processor which extracts a single double valued feature from the specified field name.
      * <p>
      * Generates features named "&lt;fieldName&gt;@value" where &lt;fieldName&gt; is the argument to this constructor if
-     * {@code onlyFieldName} is false and "&lt;fieldName&gt;" otherwise.
+     * {@code onlyFieldName} is false, otherwise generates features named "&lt;fieldName&gt;".
      * Does not throw an exception if the value failed to parse.
      * @param fieldName The field name to read.
      * @param onlyFieldName Only use the field name as the feature name.
--- a/Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
+++ b/Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
@@ -41,6 +41,12 @@ public class DoubleFieldProcessor implements FieldProcessor {
     @Config(mandatory = true,description="The field name to read.")
     private String fieldName;
 
+    @Config(description="Emit a feature using just the field name.")
+    private boolean onlyFieldName;
+
+    @Config(description="Throw NumberFormatException if the value failed to parse.")
+    private boolean throwOnInvalid;
+
     /**
      * For olcut.
      */
@@ -48,10 +54,35 @@ public class DoubleFieldProcessor implements FieldProcessor {
 
     /**
      * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * <p>
+     * Generates features named "fieldName@value", and does not throw an exception if the value failed to parse.
      * @param fieldName The field name to read.
      */
     public DoubleFieldProcessor(String fieldName) {
+        this(fieldName,false);
+    }
+
+    /**
+     * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * <p>
+     * Generates features named "fieldName@value", and does not throw an exception if the value failed to parse.
+     * @param fieldName The field name to read.
+     * @param onlyFieldName Only use the field name as the feature name.
+     */
+    public DoubleFieldProcessor(String fieldName, boolean onlyFieldName) {
+        this(fieldName,onlyFieldName,false);
+    }
+
+    /**
+     * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * @param fieldName The field name to read.
+     * @param onlyFieldName Only use the field name as the feature name.
+     * @param throwOnInvalid Throw NumberFormatException if the value failed to parse.
+     */
+    public DoubleFieldProcessor(String fieldName, boolean onlyFieldName, boolean throwOnInvalid) {
         this.fieldName = fieldName;
+        this.onlyFieldName = onlyFieldName;
+        this.throwOnInvalid = throwOnInvalid;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
```diff
@@ -41,17 +41,48 @@
...
+
     /**
      * For olcut.
      */
     private DoubleFieldProcessor() {}
 
     /**
      * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * <p>
+     * Generates features named "fieldName@value", and does not throw an exception if the value failed to parse.
```
JackSullivan, line 58: It's not 100% clear on first reading that `fieldName` is the name of the field while `value` is literally the string `value`. Is there a way to make this clearer? Maybe make `fieldName` a reference in the javadoc?

Data/src/main/java/org/tribuo/data/csv/CSVLoader.java
```diff
@@ -258,110 +255,65 @@ public CSVLoader(OutputFactory<T> outputFactory) {
...
-        //
-        // Validate that all the expected responses are included in the given header fields
-        Map<String, Boolean> responsesFound = new HashMap<>();
-        for (String response : responseNames) {
-            responsesFound.put(response, false);
-        }
-        for (String header : headers) {
-            if (responseNames.contains(header)) {
-                responsesFound.put(header, true);
+        if (!headers.containsAll(responseNames)) {
```
JackSullivan, line 284: It looks like this is to mimic the existing behavior, but is it desirable from a UX standpoint to have the user enter the response columns in both responses and headers? I feel like the intuition of keeping features and responses separate is strong and that we would want to support that.

Craigacp: The headers are used if the CSV file doesn't have headers, and so you need the column ordering.

Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
Craigacp: Sure.

Data/src/main/java/org/tribuo/data/csv/CSVLoader.java
JackSullivan: That makes sense.

Data/src/main/java/org/tribuo/data/csv/CSVLoader.java
```diff
@@ -220,7 +217,7 @@ public CSVLoader(OutputFactory<T> outputFactory) {
      * @return A datasource containing the csv data.
      * @throws IOException If the disk read failed.
      */
-    public ListDataSource<T> loadDataSource(Path csvPath, Set<String> responseNames) throws IOException {
+    public DataSource<T> loadDataSource(Path csvPath, Set<String> responseNames) throws IOException {
```
JackSullivan, line 220: You mention it in the commit message but not here, we should document that whatever order these responseNames are in will be used in the generated responses, so users should use an ordered set.

JackSullivan (commented):
Looks good apart from the documentation changes mentioned.

Data/src/main/java/org/tribuo/data/csv/CSVLoader.java
Craigacp: Ok.

Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
```diff
@@ -41,17 +41,54 @@
...
      */
     public DoubleFieldProcessor(String fieldName) {
+        this(fieldName,false);
+    }
+
+    /**
+     * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * <p>
+     * Generates features named "&lt;fieldName&gt;@value" where &lt;fieldName&gt; is the argument to this constructor if
+     * {@code onlyFieldName} is false and "&lt;fieldName&gt;" otherwise.
```
JackSullivan, line 70:
```suggestion
     * {@code onlyFieldName} is false, otherwise generates features named "&lt;fieldName&gt;".
```

JackSullivan (dismissed):
Looks good to me

Data/src/main/java/org/tribuo/data/columnar/processors/field/DoubleFieldProcessor.java
```diff
@@ -41,17 +41,54 @@
...
+     */
+    public DoubleFieldProcessor(String fieldName, boolean onlyFieldName) {
+        this(fieldName,onlyFieldName,false);
+    }
+
+    /**
+     * Constructs a field processor which extracts a single double valued feature from the specified field name.
+     * <p>
+     * Generates features named "&lt;fieldName&gt;@value" where &lt;fieldName&gt; is the argument to this constructor if
+     * {@code onlyFieldName} is false and "&lt;fieldName&gt;" otherwise.
```
JackSullivan, line 83:
```suggestion
     * {@code onlyFieldName} is false, otherwise generates features named "&lt;fieldName&gt;".
```

JackSullivan (approved):

PR #154, opened 2021-07-27
Adds ONNX export support to Tribuo's LinearSGDModels

Craigacp:
### Description
This adds support for exporting LinearSGDModels (classification, regression and multi-label) to ONNX format. The ONNX support will grow over the next few PRs as we add support for other Tribuo models, though it's likely that XGBoost and TensorFlow models will require the use of the Python converter packages to emit ONNX format.

This involves several changes:
- tribuo-multilabel-core no longer depends on tribuo-classification-sgd at test time, instead the dummy classifiers from tribuo-classification-core are used in the tests. This breaks a circular dependency between tribuo-onnx, tribuo-multilabel-core and tribuo-classification-sgd.
- tribuo-onnx now has a `MultiLabelTransformer` to allow the loading of multi-label ONNX models.
- tribuo-core now depends on protobuf and has the OnnxMl java file compiled from ONNX v1.9.0's protobuf in ai.onnx.proto. 
- The ONNX export support lives in org.tribuo.onnx in tribuo-core, and is focused around ONNXOperators which allows the construction of ONNX NodeProtos for the specified operations.
- In tribuo-common-sgd some helpers have been added to `AbstractLinearSGDModel`. This might move to `AbstractSGDModel` when we add ONNX support to factorization machines.
- Finally each `LinearSGDModel` now implements `ONNXExportable` which adds methods to create a `ModelProto` which encapsulates the whole model, a `GraphProto` which represents the model computation, and a save method which can write out an ONNX file.

As we expand the coverage to ensemble methods the `ONNXExportable` interface might change slightly, and the specific naming of the input and output nodes is likely to change (which might induce some method signature changes). I'd like to land this chunk first though as a single PR for the whole of Tribuo will be far too large to review.

One further thing that I'm considering is storing the provenance in a machine readable format inside the ONNX model, and having `ONNXExternalModel` expose that provenance, but I don't want to have everything depend on Jackson and protobuf so the current JSON based provenance string is not ideal.

### Motivation
We'd like to export Tribuo models for use in other environments, and ONNX is a popular interchange format with a friendly license that we already support model import from.


Core/src/main/java/org/tribuo/onnx/ONNXAttribute.java
```diff
@@ -0,0 +1,145 @@
...
+    public OnnxMl.AttributeProto.AttributeType getType() {
+        return type;
+    }
+
+    /**
+     * Builds the attribute proto using the supplied value.
+     * <p>
+     * Throws {@link IllegalArgumentException} if the value type does not match the expected type, and
+     * throws {@link UnsupportedOperationException} if the value type is not supported. Currently supported
+     * types are primitives, strings and arrays of primitives and strings.
```
JackSullivan, line 66: How do `long` and `double` work here? Are they unsupported by ONNX?

Craigacp: I think I could set them as `Tensor` which supports that. However this appears to be how pytorch encodes the values, as the protobufs I was generating are very similar.

In general ONNX's support of `double` is pretty lacking, to the point where I only bother to support `float` in Tribuo's ONNX Runtime interface, and I downcast everything to `float` on export.

Core/src/main/java/org/tribuo/onnx/ONNXOperators.java
```diff
@@ -0,0 +1,196 @@
...
+        if (inputs.length != numInputs) {
+            throw new IllegalArgumentException("Expected " + numInputs + " inputs, but received " + inputs.length);
+        }
+        if (outputs.length != numOutputs) {
+            throw new IllegalArgumentException("Expected " + numOutputs + " outputs, but received " + outputs.length);
+        }
+        if (attributeValues.size() > attributes.size()) {
+            throw new IllegalArgumentException("Found more attributes than expected, received " + attributeValues.size() + ", expected at most " + attributes.size());
+        }
+        if (!attributes.keySet().containsAll(attributeValues.keySet())) {
```
JackSullivan, line 162: Do we want to warn if `attributeValues` has an unrecognized key?

Craigacp: Rather than throw an exception? My intention was that this check would throw if there was an unrecognized value.

I'd consider calling the build function with an unrecognized attribute to be a programming error which should be fixed, rather than something that we'd warn at runtime over.

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/MultiLabelTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+ * output.
+ */
+public class MultiLabelTransformer implements OutputTransformer<MultiLabel> {
+    private static final long serialVersionUID = 1L;
+    private static final Logger logger = Logger.getLogger(MultiLabelTransformer.class.getName());
+
+    /**
+     * The threshold for conversion into a label.
+     */
+    public static final double THRESHOLD = 0.5;
```
JackSullivan, line 55: Should this be something that the user can adjust?

Craigacp: Probably. It's fixed at 0.5 in the binary cross entropy loss in LinearSGDModel, but this should be able to load non-probabilistic models too. I think it forces the `generateProbabilities` flag to true so that'll need to change as well.

Math/src/main/java/org/tribuo/math/util/VectorNormalizer.java
```diff
@@ -43,4 +48,21 @@ default public void normalizeInPlace(double[] input) {
...
+     * their models to ONNX. It is recommended that this method is overridden to
+     * support ONNX export, and in a future version of Tribuo this default implementation
+     * will be removed.
+     * @param context The ONNX context object for name generation.
+     * @param input The name of the input to normalize.
+     * @param output The name of the normalized output.
+     * @return A list of node protos representing the normalization operation.
+     */
+    default public List<OnnxMl.NodeProto> exportNormalizer(ONNXContext context, String input, String output) {
+        return Collections.emptyList();
```
JackSullivan, line 65: Since it is never correct to call this method directly, it should at least log a warning.

Craigacp: Sure.

Core/src/main/java/org/tribuo/onnx/ONNXOperators.java
JackSullivan: Sorry, I mean the other way around - if there are attributes defined by the operation that aren't provided by `attributeValues`. I guess some of them are required and others probably have implicit defaults, but there isn't a way to know that currently.

Craigacp: I can extend the attribute type with if it's mandatory or not and then add that check. Whether an attribute is mandatory depends on the op set version number and Tribuo is going to pin to a specific version for a while.

Core/src/main/java/org/tribuo/onnx/ONNXAttribute.java
JackSullivan: Ok, in that case maybe we should document that `long` and `double` will be downcast then, so that when it happens people will be less surprised.

Craigacp: Sure.

Core/src/main/java/org/tribuo/onnx/ONNXOperators.java
JackSullivan: I think that might be worth doing, though I think you have a better sense than I of whether or not the added overhead is worth it in improved UX. My instinct upon reflection is that it might not be - most users won't be using these methods, and the error will end up getting thrown at runtime anyway.

Craigacp: These classes are only for people extending Tribuo, not for people using it, so I'm not as worried about the amount of hand holding. I should properly document it, and it might be nice to mark the attributes so it's easy to see them in the enum, but I'm not sure if the code overhead is worthwhile.

Craigacp: I documented and added the check, it wasn't too onerous.

JackSullivan (approved):
Looks good to me.

PR #161, opened 2021-08-18
Adds a ConfigurableDataSource data generator for Clustering

Craigacp:
### Description
Adds a configurable data source for clustering demos and testing. Updated the clustering tutorial to use it.

The timing numbers are from my M1 Mac which is so fast that the multithreaded one takes roughly the same time as the single threaded one rather than being 3x faster (though it is running on 4x the data).

### Motivation
See #160 for the motivation.


jhalexand (commented):
The description for "mixingPMF" (and/or the name) could probably be a little clearer and updated both in the @Config variable's description and in the constructor.

It occurs to me that we should be checking that the variances are positive numbers both because variances are usually expressed as positive numbers and, of course, because we take the sqrt of them.

jhalexand:
It also occurs to me that the same issue, re: variances, exists in GaussianAnomolyDataSource and could be updated along with this PR.

Craigacp:
Sure, I'll fix those.

Craigacp:
I've updated the PR with your suggestions.

jhalexand (approved):
Looks good!

PR #162, opened 2021-08-19
Adds ConfigurableDataSource data generators for Classification

Craigacp:
### Description
Adds 5 data generators for 2d binary classification problems suitable for classification demos & tutorials.

They are modelled on earlier Tribuo demo datasets but updated for the current API.

### Motivation
See #160 for motivation.


jhalexand (approved):
Looks good. Math all looks right to me, though I didn't actually generate any of the data sets in reviewing this.

PR #163, opened 2021-08-21
Adds ONNX export support to the sparse linear models

Craigacp:
### Description
Adds ONNX export support to `SparseLinearModel`, which covers `SLMTrainer`, `LARSTrainer`, `LARSLassoTrainer` and `ElasticNetCDTrainer`. Added a helper method to `ONNXUtils` which can convert a `double[]` into a float Tensor vector (unfortunately due to the packaging I can't add a `DenseVector` helper because it's not visible from tribuo-core).

Also added a check to the ONNX export tests to turn off the ORT bit on arm64 platforms as ORT doesn't have an arm64 binary in the published jar on Maven Central. The tests do pass on arm64 when ORT is compiled on arm64, so if MS add the binary to Maven Central we can remove this restriction.

Finally it also fixes a bug in `RegressionFactory.constructInfoForExternalModel` which previously couldn't construct infos for multidimensional regressions. This is copied across from the larger regression id fixes, so we'll backport that changeset to 4.1 rather than try to extract it from this one.

### Motivation
Expanding the set of ONNX exportable models.

### DIFF
--- a/Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
+++ b/Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
@@ -67,7 +67,7 @@ public abstract class ONNXUtils {
         ByteBuffer buffer = ByteBuffer.allocate(parameters.length*4).order(ByteOrder.LITTLE_ENDIAN);
         FloatBuffer floatBuffer = buffer.asFloatBuffer();
         for (int i = 0; i < parameters.length; i++) {
-            floatBuffer.put((float)parameters.length);
+            floatBuffer.put((float)parameters[i]);
         }
         floatBuffer.rewind();
         arrBuilder.setRawData(ByteString.copyFrom(buffer));
--- a/Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
+++ b/Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
@@ -17,6 +17,11 @@
 package org.tribuo.onnx;
 
 import ai.onnx.proto.OnnxMl;
+import com.google.protobuf.ByteString;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.FloatBuffer;
 
 /**
  * Helper functions for building ONNX protos.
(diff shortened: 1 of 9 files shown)


Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
```diff
@@ -44,4 +49,29 @@ private ONNXUtils() {}
 
         return builder.build();
     }
+
+    /**
+     * Builds a TensorProto containing the array.
+     * <p>
+     * Downcasts the doubles into floats as ONNX's fp64 support is poor compared to fp32.
```
JackSullivan, line 56: Do we want to future-proof this a bit by supporting both fp64 and fp32 now, but defaulting to downcasting? That way we won't need to change the interface if support gets better or we care more about it in the future?

Craigacp: I can add an overload with a boolean flag and have this one call it with `downcast` to true.

Regression/SLM/src/main/java/org/tribuo/regression/slm/SparseLinearModel.java
```diff
@@ -189,14 +201,149 @@ protected DimensionTuple scoreDimension(int dimensionIdx, SparseVector features)
...
+        builder.setIrVersion(6);
+        return builder.build();
+    }
+
+    /**
+     * Builds a TensorProto containing the model weights.
+     * @param context The naming context.
+     * @return The weight TensorProto.
+     */
+    protected OnnxMl.TensorProto weightBuilder(ONNXContext context) {
```
JackSullivan, line 241: shouldn't these be factored out into `ONNXHelpers` or something - they're repeated almost verbatim from `AbstractLinearSGDModel`, and expect this won't be the last time we want to convert `DenseVector[]`/`DenseMatrix`/`DenseVector` into a `OnnxMl.TensorProto`.

Regression/SLM/src/main/java/org/tribuo/regression/slm/SparseLinearModel.java
```diff
@@ -189,14 +201,149 @@ protected DimensionTuple scoreDimension(int dimensionIdx, SparseVector features)
...
+
+        // Add feature and output means
+        double[] xMean = bias ? Arrays.copyOf(featureMeans.toArray(),featureIDMap.size()) : featureMeans.toArray();
+        OnnxMl.TensorProto featureMeanProto = ONNXUtils.arrayBuilder(context, "feature_mean",xMean);
+        graphBuilder.addInitializer(featureMeanProto);
+        OnnxMl.TensorProto outputMeanProto = ONNXUtils.arrayBuilder(context,"y_mean",yMean);
+        graphBuilder.addInitializer(outputMeanProto);
+
+        // Add feature and output variances
+        double[] xVariance = bias ? Arrays.copyOf(featureVariance.toArray(),featureIDMap.size()) : featureVariance.toArray();
```
JackSullivan, line 316: How does ONNX handle the potential `NaN` values here?

Regression/SLM/src/main/java/org/tribuo/regression/slm/SparseLinearModel.java
Craigacp: I could add another helper class into `tribuo-math` or add ONNX exporting functionality onto `Tensor` itself, but it can't live on `ONNXHelpers` as `DenseVector` et al aren't visible in `tribuo-core`. Unfortunately the SLM and SGD uses are all slightly different too, as SGD doesn't have a `DenseVector` for the biases it's packed onto the weight matrix.

Regression/SLM/src/main/java/org/tribuo/regression/slm/SparseLinearModel.java
Craigacp: The variances are initialized to 1 if they are unused, and if you hit a zero variance feature something else is likely to have blown up before you got to the model part. I don't actually know what ONNX would do if there was a NaN here, I'd expect it to propagate it.

JackSullivan (changes requested):
Looks good apart from support for fp64 in `ONNXUtils`

JackSullivan (approved):
Looks good to me

Regression/SLM/src/main/java/org/tribuo/regression/slm/SparseLinearModel.java
Craigacp: We're going to leave this until we have a third use-case and then see if we can extract common functionality from all three.

Core/src/main/java/org/tribuo/onnx/ONNXUtils.java
Craigacp: Done.

PR #167, opened 2021-08-31
Fixing a bug where MultiLabelConfusionMatrix swapped FP for FN

Craigacp:
### Description
The multilabel evaluation package swapped the FPs for FNs when accessing them, causing many of the metrics to be permuted incorrectly producing incorrect values for precisions, recalls and the evaluation.toString().

### Motivation
It's best if the evaluations report the right numbers in the right places.


pogren (commented):
It looks like the class-level javadoc comments for MultiLabelConfusionMatrix still have the indices backwards.

pogren (approved):
This is a straightfoward bug fix that corrects the indices used for retrieving FPs and FNs from the underlying matrix used in MultiLabelConfusionMatrix (i.e. the "fp" and "fn" methods).  Now the values used match the way the data is entered in the tabulate method.  The only other changes are minor javadoc edits and fixing the assertion values in the unit test.

PR #168, opened 2021-08-31
Updating the multi-label tutorial after the evaluation bug fix

Craigacp:
### Description
Updates the evaluations and text in the new multi-label tutorial after the fix for #167.


pogren (approved):
Mostly just updates the notebook to reflect updated output after the evaluation bug fix

PR #170, opened 2021-09-02
Bumping LibSVM version

Craigacp:
### Description
Bump LibSVM to 3.25.

### Motivation
Moving to LibSVM's latest version for compliance reasons (there are no Java fixes in this version vs 3.24 as far as I can tell).


jhalexand (approved):
Done

PR #171, opened 2021-09-03
CSVDataSource should persist the headers in the provenance

Craigacp:
### Description
The headers were set after the provenance was constructed, so the provenance only recorded headers set via configuration. This PR fixes that bug and adds a test for the behaviour. It also makes the CSVDataSource constructor which accepts headers public so it can be used by developers. An additional public constructor which accepts a header and a Path was added to maintain parity.

### Motivation
When using CSVLoader on a file without headers it should persist the headers in the provenance so the configuration is recoverable. Also the CSVDataSource constructor which accepts headers is generally useful if people wish to construct a CSVDataSource for csvs without headers manually.


JackSullivan (approved):
Looks good to me

PR #172, opened 2021-09-06
LibLinear and LibSVM have unmanaged global RNGs

Craigacp:
### Description
This change fixes Tribuo's LibLinear and LibSVM interfaces so they use trackable RNG states. 

LibLinear has a protected global RNG with a reset method (to a fixed seed), so now Tribuo sequentialises all LibLinear training runs under a lock on `LibLinearTrainer.class` and resets the seed before each call to `Linear.train`. There's a new reproducibility test in the classification LibLinear module which fails without these changes. This fix may change depending on the resolution of https://github.com/bwaldvogel/liblinear-java/issues/40.

LibSVM has a public global RNG so we use Tribuo's standard RNG idiom (a per train call `localRNG` split from the trainer's instance of `SplittableRandom`), LibSVM's global RNG's seed is set to a draw from `localRNG` and all LibSVM training runs under a lock on `LibSVMTrainer.class`. There's a new reproducibility test in the classification LibSVM module which fails without these changes. Unfortunately it's quite a chatty test as LibSVM prints a lot during this run.

### Motivation
We need Tribuo models to be reproducible.

### DIFF
--- a/AnomalyDetection/LibLinear/src/main/java/org/tribuo/anomaly/liblinear/LibLinearAnomalyTrainer.java
+++ b/AnomalyDetection/LibLinear/src/main/java/org/tribuo/anomaly/liblinear/LibLinearAnomalyTrainer.java
@@ -41,7 +41,9 @@ import java.util.logging.Logger;
 
 /**
  * A {@link Trainer} which wraps a liblinear-java anomaly detection trainer using a one-class SVM.
- *
+ * <p>
+ * Note the train method is synchronized on {@code LibLinearTrainer.class} due to a global RNG in liblinear-java.
+ * <p>
  * See:
  * <pre>
  * Fan RE, Chang KW, Hsieh CJ, Wang XR, Lin CJ.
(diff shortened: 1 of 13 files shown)


pogren (approved):
The train methods in the trainers for LibLinear and LibSVM call Linear.resetRandom() svm.rand.setSeed(localRNG.nextLong()), respectively. The latter looks a little suspect at first glance but note that different instances of LibSVMTrainer will have identical localRNG instances.  The respective train methods also synchronize on LibLinearTrainer.class and LibSVMTrainer.class.   A couple of unit tests demonstrate that subsequent runs of the train method for both produce the same model parameters.  

The trade-off between locking on classes Tribuo does not control (not done) and the risk of someone using Liblinear in the same jvm in a non-synchronized way and mucking things up is documented.  

SVMAnomalyType.isClassification and .isAnomaly had the boolean return values flipped.  oops!  Corrected here.  

Various and sundry documentation comments are sprinkled in the code

PR #175, opened 2021-09-17
Add Javadoc for all remaining undocumented public methods and fields.

Craigacp:
### Description
Added javadoc for all undocumented public classes, methods and fields. Apart from AbstractSGDTrainer/AbstractSGDModel as those are documented in the factorization machines branch and so will be fixed then.

### Motivation
Docs are good.


jhalexand (approved):
LGMT

PR #176, opened 2021-09-17
Two small fixes for provenance.

Craigacp:
### Description
Switches around how the provenance is captured in TransformTrainer so it has the right values, and updates XGBoostExternalModel so there is a way to create a model from an in-memory XGBoost `Booster`, which required an extra constructor on `ExternalTrainerProvenance`.

### Motivation
The transform trainer was capturing the incorrect state of the inner trainer's RNG, and the XGBoost external model creator would always throw an exception if you didn't have a valid URL to a path on disk that it could hash.

### DIFF
--- a/Common/XGBoost/src/main/java/org/tribuo/common/xgboost/XGBoostExternalModel.java
+++ b/Common/XGBoost/src/main/java/org/tribuo/common/xgboost/XGBoostExternalModel.java
@@ -16,6 +16,7 @@
 
 package org.tribuo.common.xgboost;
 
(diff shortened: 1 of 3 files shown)


pogren (approved):
New constructor on ExternalTrainerProvenance takes a byte array and manages to create a working provenance using the hash value generated from the byte array, the current time, and the current working directory.  This is used by XGBoostExternalModel.createXGBoostModel so that the provenance works.  Also, some care has been given to synchronize the training and provenance construction in TransformTrainer.

PR #177, opened 2021-09-24
Fixes multidimensional regression

Craigacp:
### Description
Fixes `ImmutableRegressionInfo` so it generates the regression ids in lexicographic order. It then fixes all the models so they correctly map their dimensions through the `ImmutableRegressionInfo` so even if it's not in lexicographic order then the models produce the correct output. Finally it adds deserialization hooks to `LibLinearRegressionModel`, `LibSVMRegressionModel`, `XGBoostModel` and `SparseLinearModel`s trained using `ElasticNetCDTrainer` so that older models are rewritten on deserialization to fix the issue. Once those models have been loaded by a version of Tribuo with this fix they will work correctly in older versions (though if you then save them again in the older version they will become irretrievably corrupt as it'll lose the flag that tracks if the model has been fixed).

Separately this PR also fixes an issue where multidimensional LibSVM models are corrupted when using the standardization flag. Only the first dimension was stored correctly, the others are copies of that dimension.

Unfortunately tree based models (`TreeModel<Regresssor>`, `IndependentRegressionTreeModel` and any ensembles created using them) are more deeply corrupted and a fix would require completely rewriting the tree structure. We recommend users retrain those models in a newer version of Tribuo which contains this fix. Additionally TensorFlow based multidimensional regression models are likely impacted by this issue and should be retrained.

One further additional change in this PR is that `XGBoostModel` now correctly reports the top features on a per regression dimension basis rather than aggregating them into a single list as it did previously. This was useful for debugging the various fixes to XGBoost.

### Motivation
Multidimensional regression models in v4.0 and v4.1.0 have an incorrect mapping between ids and regression dimension names. The root cause is two-fold, due to an implicit contract which was not enforced (which is that the regression id numbers are generated using a lexicographic sort of the regression dimension names). The first cause is that when building an `ImmutableRegressionInfo` the labels were accidentally put into a non-sorted Map, and then iterated over in that map causing the ids to be hash order dependent, rather than lexicographically sorted. This shouldn't have been enough to break Tribuo, as all id mappings should go through the `ImmutableRegressionInfo` anyway. The second cause is that they didn't and several trainers/models assumed that the regression values were stored in id order in the `Regressor` object itself.

The LinearSGDModel and SparseLinearModels (aside from ones generated using the ElasticNetCDTrainer) are correct in older versions of Tribuo. LibLinear, LibSVM and XGBoost models produce the correct output, but are corrupted internally so store the models in the incorrect locations. Tree models are entirely corrupted and should be retrained from scratch.

Note single dimensional regression models are unaffected by this bug as it only occurs when there are two dimensions which could be stored out of order, and we expect those are the vast majority of trained regression models.


Regression/Core/src/main/java/org/tribuo/regression/ImmutableRegressionInfo.java
```diff
@@ -58,37 +63,67 @@ private ImmutableRegressionInfo(ImmutableRegressionInfo info) {
...
+            }
+            else if (curNames.length == 1) {
+                names[e.getValue()] = curNames[0];
             } else {
                 throw new IllegalArgumentException("Mapping must contain a single regression dimension per id, but contains " + Arrays.toString(names) + " -> " + e.getValue());
             }
         }
+        idLabelMap = new LinkedHashMap<>();
+        labelIDMap = new LinkedHashMap<>();
+        for (int i = 0; i < names.length; i++) {
```
pogren, line 120: Seems like 'names' should be sorted here but I don't think that it is.

Craigacp: No, this is the intentional escape hatch, used when the output info needs to be constructed for an external model (e.g. loading in an ONNX or TF model) which doesn't have the same invariants as Tribuo requires.

pogren: ok - I see that in the javadoc now.  thanks for the clarification.

pogren (approved):
most of the code is updated references to the dimension ids using the updated mapping, a bunch of new/updated unit tests to make sure it all works correctly, new readObject methods, and a rewrite of ImmutableRegressionInfo
some smaller changes include: fixing LinearRegressionType return values for isClassification and isRegression, changing use of LinkedList to ArrayDeque

PR #178, opened 2021-09-24
ResponseProcessor State-setting and Tests

JackSullivan:
### Description
Fixes an issue with multi-field `ResponseProcessor` instances where the end state of the produced object, when written back out into a configured/provenanced form and re-instantiated wouldn't be processed correctly by the `postConfig` call. Also added a test for roundtripping any type that is both `Configurable` and `Provenanceable` through their respective forms to ensure that the end results are structurally equal according to the `ConfigurationData#structuralEquals` method in OLCUT, and tests for this behavior in `BinaryResponseProcessor` and `FieldResponseProcessor`.

This change depends on OLCUT v5.2.0, and thus requires updating Tribuo to depend on that OLCUT version.


Data/src/main/java/org/tribuo/data/columnar/processors/response/FieldResponseProcessor.java
```diff
@@ -64,20 +64,23 @@
     public void postConfig() {
         if (fieldName != null && fieldNames != null) {
             throw new PropertyException(configName, "fieldName, FieldNames", "only one of fieldName or fieldNames can be populated");
-        } else if (fieldNames != null) {
-            if(defaultValue != null) {
-                defaultValues = defaultValues == null ? Collections.nCopies(fieldNames.size(), defaultValue) : defaultValues;
+        } else if (fieldNames != null) { // multiple fields
+            if (defaultValues != null && defaultValues.size() == fieldNames.size()) { // check first for multiple defaults
+                // this is the default case, nothing needs to be done
```
Craigacp, line 69: Can we flip this around to get rid of the empty if branch? Maybe start with `defaultValues == null && defaultValue == null` to catch the else case, and then have the `defaultValue != null` one, then `defaultValues != null && defaultValues.size() != fieldNames.size()` to throw `PropertyException`.

Craigacp (changes requested):
I think the if/else in FieldResponseProcessor could be simplified a bit, otherwise it looks fine.

PR #179, opened 2021-09-27
Factorization machines

Craigacp:
### Description
Adds pairwise Factorization Machines trained using SGD, for multi-class classification, multi-label classification and regression. Factorization Machines can also be exported as ONNX models, though the translation uses dense tensors due to a lack of support for Sparse Tensors in ONNX Runtime (we might revise this as ORT's support improves). There are few implementations of multi-class factorization machines, but the code uses the obvious extension to multiple dimensions (add a per dimension bias, linear weight vector and feature embedding matrix). Due to the per output dimension embedding matrix this might not scale to large numbers of output dimensions particularly well, in such cases an MLP might be preferable.

Compared with other libraries for factorization machines the Tribuo implementation is missing l2 regularisation (which could be added by using the `ShrinkingTensor` but we didn't have time to add the necessary support), though adds support for using any gradient optimiser instead of just SGD or Adagrad. We plan to add regularisation support at a later date, along with adding l1 and l2 regularisation to `LinearSGDTrainer`.

In addition to the Factorization Machines there are:
- several cleanups to the SGD packages, including adding names to exported ONNX graphs (this is mandatory according to the ONNX spec, but ORT doesn't seem to check it).
- a few additions to the Math package.
- `DataOptions` can now standardize the features on load.
- `RegressionInfo` has methods to access the mean, std deviation and other computed statistics. These are re-computed on deserialization rather than being stored in the serialized object.
- several cleanups for the ONNX support including extra overloads for `ONNXOperators.build` and optional inputs support.

### Motivation
Factorization Machines are a powerful prediction technique which allow pairwise feature interactions via a feature embedding matrix. This makes them more powerful than logistic regression, but not as complex as an MLP, providing a nice intermediate model which scales well under sparsity (as if the features aren't present they don't appear in the factorized representation).

### Paper reference
S. Rendle, Factorization Machines, 2010 IEEE Conference on Data Mining.


Common/SGD/src/main/java/org/tribuo/common/sgd/AbstractSGDTrainer.java
```diff
@@ -158,17 +161,23 @@ public V train(Dataset<T> examples, Map<String, Provenance> runProvenance) {
...
             } else {
                 sgdFeatures[n] = SparseVector.createSparseVector(example, featureIDMap, addBias);
             }
             sgdTargets[n] = getTarget(outputIDInfo,example.getOutput());
+            featureSize += sgdFeatures[n].numActiveElements();
             n++;
         }
         logger.info(String.format("Training SGD model with %d examples", n));
+        logger.fine("Average feature size = " + featureSize / (double)n);
+        logger.fine("Dense count = " + denseCount);
```
JackSullivan, line 180: These log messages aren't clear to someone who hasn't read this code. Maybe something like: 
```suggestion
        logger.fine("Mean number of active features = " + featureSize / (double)n);
        logger.fine("Number of dense examples = " + denseCount);
```

Craigacp: Sure. To be honest this could go, I was using it to track down a performance issue, but it'll probably be useful logging for the next performance issue.

Math/src/main/java/org/tribuo/math/la/SparseVector.java
```diff
@@ -459,13 +471,40 @@ public void hadamardProductInPlace(Tensor other, DoubleUnaryOperator f) {
         }
     }
 
+    /**
+     * Applies a {@link ToDoubleBiFunction} elementwise to this {@link SGDVector}.
```
JackSullivan, line 475:
```suggestion
     * Applies a {@link DoubleUnaryOperator} elementwise to this {@link SGDVector}.
```

Math/src/main/java/org/tribuo/math/la/SparseVector.java
```diff
@@ -459,13 +471,40 @@ public void hadamardProductInPlace(Tensor other, DoubleUnaryOperator f) {
...
+     * <p>
+     * The first argument to the function is the index, the second argument is the current value.
```
JackSullivan, line 477:
```suggestion
     * Applies a {@link DoubleUnaryOperator} elementwise to this {@link SGDVector}.
```

Craigacp: Ooops, thanks.

Regression/SGD/src/main/java/org/tribuo/regression/sgd/TrainTest.java
```diff
@@ -86,22 +86,22 @@ public String getOptionsDescription() {
         /**
          * Number of SGD epochs. Defaults to 5.
          */
-        @Option(charName = 'i', longName = "epochs", usage = "Number of SGD epochs. Defaults to 5.")
+        @Option(charName = 'i', longName = "epochs", usage = "Number of SGD epochs.")
```
JackSullivan, line 89: Are these defaults now displayed in a user-visible way elsewhere? I know that we don't consistently record them in usage strings, but I don't think that means that we should remove them from places where they do appear - they're important for users who aren't looking at the source code.

Craigacp: Yes, the options system prints them when you ask for usage:
```
CARTClassificationOptions

Description: Options for decision/classification trees.
Char Long Name                  Type                                             Default Usage
     cart-impurity              enum - {GINI, ENTROPY}                           GINI    Impurity measure to use. Defaults to GINI.
     cart-max-depth             int                                              6       Maximum depth in the decision tree.
     cart-min-child-weight      float                                            5.0     Minimum child weight.
     cart-min-impurity-decrease float                                            0.0     Minimum impurity decrease.
     cart-print-tree            boolean                                          false   Prints the decision tree.
     cart-random-split          boolean                                          false   Whether to choose split points for features at random.
     cart-seed                  long                                             12345   RNG seed.
     cart-split-fraction        float                                            1.0     Fraction of features in split.
     cart-tree-algorithm        enum - {CART}                                    CART    Tree algorithm to use (options are CART).
```

PR #181, opened 2021-10-05
Normalizes the URL created from paths in LibSVMDataSource

Craigacp:
### Description
This change normalizes the path before it is converted into a URL in LibSVMDataSource. Adds a test for the fixed behaviour.

### Motivation
`LibSVMDataSource.postConfig` checks to see if the URL and path are the same before throwing an exception if both are set. The `FileProvenance` constructor normalizes and absolutes the path before it is created, however the original codepath in `LibSVMDataSource` did not do this to the URL which is created. This means if the path has a relative element in it like `.` or `..` then the URL created retains the relative element and the path has it normalized out, meaning the two values are not equal and an exception is thrown when the configuration is instantiated from the provenance of a `LibSVMDataSource`.


JackSullivan (approved):
Looks good to me

PR #185, opened 2021-10-19
Addition of a Reproducibility Framework 

jwons:
### Description
Adding a reproducibility framework that uses Tribuo's self-describing models and OLCUT to re-train a given model. It also exposes a configuration manager allowing users to modify the configuration before re-training. This framework also can create a diff between two model provenance objects, for the purpose of comparing a model to its reproduced version.   

The framework is a new module within Tribuo that is targeted at Java 16. The module contains a new class, `ReproUtil`.
A `ReproUtil` must be instantiated with either a `Model` or `ModelProvenance` object. Once instantiated, users can access the OLCUT configuration manager using a getter method. Two methods on the `ReproUtil` object will re-train a model. 
Either `reproduceFromProvennace` or `reproduceFromModel`. The  `reproduceFromModel` will only execute if the user passed a `Model` object upon initial instantiation. Both methods will retrain a model from the configuration stored within a the model's provenance; however, reproducing from a model will additionally compare metadata not stored in the provenance (feature and output domains).

### Motivation
This is a new feature that allows automatic reproducibility for any Tribuo-trained model. This framework helps increase the reproducibility of data science conducted using Tribuo, and can act as a basis for an experimental framework since the framework control all aspects of a model's configuration.

### DIFF
--- a/Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
+++ b/Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
@@ -436,7 +436,7 @@ public class ReproUtil {
     }
 
     public record FeatureDiff (Set<String> originalFeatures, Set<String> reproducedFeatures){}
-    public record OutputDiff<T extends Output>(Set<T> originalOutput, Set<T> reproducedOutput){}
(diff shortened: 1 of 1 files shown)


Craigacp:
So it looks like the tests directly pass on Java 16 which is good, and I'll poke at the CI a little to figure that out.

PR #186, opened 2021-10-22
ONNX ensemble support

Craigacp:
### Description
Adds support for exporting Tribuo's `WeightedEnsembleModel` to ONNX. It includes support for nested ensembles, and required a rewrite of the `ONNXExportable` interface. Also this PR bumps the ONNX opset to 13. In addition to the ONNX changes, export support is added to all the implementations of `EnsembleCombiner` in Tribuo.

### Motivation
We'd like to export more models to ONNX format.


Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
```diff
@@ -216,4 +224,89 @@ public WeightedEnsembleModel(String name, EnsembleModelProvenance provenance, Im
...
+        OnnxMl.NodeProto concat = ONNXOperators.CONCAT.build(context,unsqueezeOutputs,context.generateUniqueName("ensemble_concat"), Collections.singletonMap("axis",2));
+        context.addNode(concat);
+
+        // Write in weights initializer
+        OnnxMl.TensorProto weightTensor = ONNXUtils.arrayBuilder(context,"ensemble_weights",weights);
+        context.addInitializer(weightTensor);
+
+        // Write in ensemble combiner
+        List<OnnxMl.NodeProto> combinerNodes = combiner.exportCombiner(context,concat.getOutput(0),outputName,weightTensor.getName());
+        if (combinerNodes.isEmpty()) {
```
JackSullivan, line 307: This is sort of an ugly hack, as that error message could potentially hide a different (also presumably erroneous) reason that the combiner is empty.

Core/src/main/java/org/tribuo/onnx/ONNXOperators.java
```diff
@@ -75,6 +78,26 @@
      * Element-wise exponentiation with broadcasting.
      */
     POW("Pow",2,1),
+    /**
+     * Hardmax(element in input, axis) = 1 if the element is the first maximum value along the specified axis, 0 otherwise.
```
JackSullivan, line 82: Unless 'first' here refers to something specific.
```suggestion
     * Hardmax(element in input, axis) = 1 if the element is the maximum value along the specified axis, 0 otherwise.
```

Core/src/main/java/org/tribuo/ensemble/WeightedEnsembleModel.java
Craigacp: The default method on `EnsembleCombiner` returns an empty list, and it's the same idiom as we used in the `VectorNormalizer`s in `LinearSGDModel`. I can update the docs for `exportCombiner` to say it only returns an empty list if the method isn't implemented, otherwise it should return the identity operation. And then say other erroneous cases are handled by throwing a runtime exception.

Core/src/main/java/org/tribuo/onnx/ONNXOperators.java
Craigacp: Actually that's from the ONNX spec for the operator, it only sets the value to 1 for the first maximum element - https://github.com/onnx/onnx/blob/master/docs/Operators.md#Hardmax.

Craigacp:
Rebased to fix the merge conflict introduced by merging the reproducibility PR.

JackSullivan (dismissed):
Looks good to me

JackSullivan (approved):
Looks even better!

PR #189, opened 2021-10-28
Reproducibility generics cleanup

Craigacp:
### Description
Cleaning up the generics in the reproducibility so it compiles without warnings. It adds an output type to the `ReproUtil` as the target type inference used previously could infer the incorrect type from the calling environment.

Also added support for extracting the provenance from an `ONNXExternalModel` if it contains a Tribuo provenance field in the ONNX model.

### Motivation
Compiler warnings are bad.


Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
```diff
@@ -151,33 +188,32 @@ private ReproUtil(ModelProvenance provenance, Model<?> originalModel) {
...
-     * @return A {@link DataSource} object that was used to load data from te original model
-     * @throws IllegalArgumentException
+     * This method takes a class object and will recover that datasource from the {@link ConfigurationManager}.
+     * <p>
+     * Throws {@link IllegalStateException} if the data source cannot be instantiated.
+     * @param dataSourceClass A configurable class that this method will search for in the configuration manager.
+     * @return A {@link DataSource} object that was used to load data from te original model.
      */
-    private <T extends Output<T>> DataSource<T> getDatasourceFromCM(Class<? extends Configurable> dataSourceClass){
+    private DataSource<T> getDatasourceFromCM(Class<? extends Configurable> dataSourceClass){
```
JackSullivan, line 216: Shouldn't this be?
```suggestion
    private <D extends DataSource<T>> D getDatasourceFromCM(Class<D> dataSourceClass){
```

Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
```diff
@@ -196,52 +232,41 @@ private ReproUtil(ModelProvenance provenance, Model<?> originalModel) {
...
+        if (sourceName == null){
+            throw new IllegalStateException("Failed to find source with expected class " + dataSourceClass);
         }
 
-        // Throw error if the source cannot be instantiated.
-        if (dataSource == null) {
-            throw new IllegalArgumentException("The provided provenance has no data source");
+        // Type is restricted by source construction
+        @SuppressWarnings("unchecked")
+        DataSource<T> dataSource = (DataSource<T>) cm.lookup(sourceName);
```
JackSullivan, line 241: As above
```suggestion
        D dataSource = (D) cm.lookup(sourceName);
```

Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
Craigacp: I sharpened it to `<D extends ConfigurableDataSource<T>>`, as `DataSource` is not necessarily a subclass of `Configurable`.

Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
Craigacp: As we have the class, I used `dataSourceClass.cast(cm.lookup(sourceName))` which allowed me to drop the warning suppression.

Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
```diff
@@ -299,30 +329,33 @@ private ReproUtil(ModelProvenance provenance, Model<?> originalModel) {
     }
 
     /**
-     * Return a {@link Dataset} used when a model was trained
-     * @param <T> The {@link Dataset} type
-     * @return A new {@link Dataset}
-     * @throws IllegalStateException
-     * @throws ClassNotFoundException
+     * Return a {@link Dataset} used when a model was trained.
```
JackSullivan, line 332: Do we want to note here that we currently only handle train/test for `TrainTestSplitter`and not any kind of cross-validation?

Craigacp: Probably. The information is in the provenance, but we'd need a special case for `DatasetView` in `recoverDataset`.

I'm not sure how many models there are trained using cross validation though. While we support it the API is in need of some attention.

JackSullivan (changes requested):
We should document that this doesn't currently support `TrainTestSplitter` with validation and `DatasetView`, but otherwise looks good.

Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
Craigacp: Fixed.

JackSullivan (approved):
Looks good to me

PR #191, opened 2021-11-02
ONNX export support for LibLinear and LibSVM

Craigacp:
### Description
Adds ONNX export support for LibLinear and LibSVM classification and regression models. It doesn't add anomaly detection export support, we'll leave that till a future release.

The LibSVM classification export depends on the exact behaviour of ONNX Runtime as the ONNX spec for SVMClassifier is incorrect - https://github.com/microsoft/onnxruntime/issues/9429.

### Motivation
We'd like to export models to ONNX. Also LibSVM uses an operator from the ONNX-ML set and so is a good test of the support.

### DIFF
--- a/Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
+++ b/Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
@@ -94,7 +94,7 @@ public final class LabelOneVOneTransformer extends LabelTransformer {
                     throw new IllegalArgumentException("Expected the first element to be a float OnnxTensor, found " + inputs.get(0));
                 }
             } else if (inputs.size() == 2) {
-                if (inputs.get(1) instanceof OnnxTensor) {
+                if (inputs.get(0) instanceof OnnxTensor && inputs.get(1) instanceof OnnxTensor) {
                     OnnxTensor outputLabels = (OnnxTensor) inputs.get(0);
                     OnnxTensor outputScores = (OnnxTensor) inputs.get(1);
                     if (outputScores.getInfo().type == OnnxJavaType.FLOAT) {
--- a/Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
+++ b/Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
@@ -28,6 +28,23 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Logger;
 
+/**
+ * Can convert an {@link OnnxValue} into a {@link org.tribuo.Prediction} or a {@link Label}.
+ * <p>
+ * Accepts:
+ * <ul>
+ *     <li>a tuple (tensor, float tensor) - as produced by the bare ONNX ML operations (e.g., SVMClassifier).</li>
+ *     <li>a single float tensor.</li>
+ * </ul>
+ * It attempts to parse the output as if it's a vector of predictions from one-v-one classifiers for each class pair.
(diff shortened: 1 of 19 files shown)


Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public void postConfig() {
+        if (generatesProbabilities) {
+            throw new PropertyException("", "generatesProbabilities", "generatesProbabilities must not be set to true for this class.");
+        }
+    }
+
+    /**
+     * Rationalises the output of an onnx model into a standard format suitable for
```
JackSullivan, line 50: This should probably mention that it accept/expects similar input types to `LabelTransformer`

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
```diff
@@ -0,0 +1,150 @@
...
+                            throw new IllegalArgumentException("Invalid shape for the score tensor, expected shape [batchSize,(numOutputs*(numOutputs-1))/2], found " + Arrays.toString(shape));
+                        }
+                    } else {
+                        throw new IllegalArgumentException("Expected the first element to be a float OnnxTensor, found " + inputs.get(0));
+                    }
+                } else {
+                    throw new IllegalArgumentException("Expected the first element to be a float OnnxTensor, found " + inputs.get(0));
+                }
+            } else if (inputs.size() == 2) {
+                if (inputs.get(1) instanceof OnnxTensor) {
```
JackSullivan, line 97: shouldn't this be:
```suggestion
                if (inputs.get(0) instanceof OnnxTensor && inputs.get(1) instanceof OnnxTensor) {
```

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
```diff
@@ -0,0 +1,150 @@
...
+    @Override
+    protected float[][] getBatchPredictions(List<OnnxValue> inputs, ImmutableOutputInfo<Label> outputIDInfo) {
+        try {
+            if (inputs.size() == 1) {
+                // Single OnnxTensor [batchSize][(numOutputs*(numOutputs-1))/2]
+                if (inputs.get(0) instanceof OnnxTensor) {
+                    OnnxTensor outputScores = (OnnxTensor) inputs.get(0);
+                    if (outputScores.getInfo().type == OnnxJavaType.FLOAT) {
+                        long[] shape = outputScores.getInfo().getShape();
+                        if ((shape.length == 2) && (shape[1] == (outputIDInfo.size() * ((long) outputIDInfo.size() - 1) / 2))) {
```
JackSullivan, line 66: Should this have the `|| shape[1] == 2` as the below clause does?

JackSullivan (changes requested):
Looks good apart from minor changes in comments

Interop/ONNX/src/main/java/org/tribuo/interop/onnx/LabelOneVOneTransformer.java
Craigacp: Maybe. ONNX Runtime automatically unfolds the output of a SVMClassifier so it has two columns, but in general if you're building `(n * n-1) / 2` classifiers for your `n` class problem you'll end up with 1 output for binary problems, and if ONNX Runtime hasn't special cased it you'll produce a single output tensor of size `[batch_size,1]` which the other logic covers. Neither ONNX Runtime or the ONNX spec are really correct here, so I added some docs to explain that we're in an area prone to change because it's implementation dependent.

JackSullivan (approved):
Looks good to me

PR #195, opened 2021-11-28
Tensorflow 0.4.0 update

Craigacp:
### Description
Bump TF-Java to 0.4.0 from 0.3.1. TF-Java changed how initialisation works which results in a few changes in Tribuo's APIs, but our TF interface is not covered by the same stability guarantee as the rest of Tribuo.

### Motivation
TF-Java 0.3 pulls in an older version of TF which has several bugs.


jhalexand (approved):
LGTM

PR #196, opened 2021-11-29
These are the changes for an implementation of HDBSCAN*

geoffreydstewart:
### Description
An implementation of the HDBSCAN* clustering algorithm is being proposed. It is a new module, which is being added adjacent to the KMeans module.

### Motivation
This is a new feature. The HDBSCAN* clustering algorithm is a density-based algorithm. Unlike K-Means, it doesn't require that every data point is assigned to a cluster, since it identifies dense clusters. Points not assigned to a cluster are considered as outliers, or anomalies. An algorithm that can effectively find distinct groups in a dataset and identify outliers is a valuable technique. 

### Paper reference
[1] R. J. Campello, D. Moulavi, and J. Sander, "Density-based Clustering Based on Hierarchical Density Estimates," in Pacific-Asia Conference on Knowledge Discovery and Data Mining. Springer, 2013, pp. 160–172.


Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+    private static List<Integer> findProminentClusters(Map<Integer, int[]> hierarchy, List<HdbscanCluster> clusters,
+                                                      int numPoints) {
+
+        // Take the list of propagated clusters from the root cluster:
+        List<HdbscanCluster> solution = clusters.get(1).getPropagatedDescendants();
+
+        List<Integer> clusterLabels = new ArrayList<>(Collections.nCopies(numPoints, OUTLIER_NOISE_CLUSTER_LABEL));
+
+        // Store all the levels at which to find the birth points for the flat clustering:
+        TreeMap<Integer, List<Integer>> significantLevels = new TreeMap<>();
```
Craigacp, line 594: This also seems like it could be a priority queue, though it might need to be a queue of a `Pair<Integer,List<Integer>>`.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+     */
+    private static DenseVector calculateOutlierScores(double[] pointNoiseLevels, int[] pointLastClusters,
+                                                                 List<HdbscanCluster> clusters) {
+
+        int numPoints = pointNoiseLevels.length;
+        DenseVector outlierScores = new DenseVector(numPoints);
+
+        // Iterate through each point, calculating its outlier score:
+        for (int i = 0; i < numPoints; i++) {
+            double epsilon_max = clusters.get(pointLastClusters[i]).getPropagatedLowestChildDeathLevel();
```
Craigacp, line 634: Can you replace the `snake_case` with `camelCase`?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+     * @return A map of the cluster labels, and the points assigned to them.
+     */
+    private static Map<Integer, TreeMap<Double, Integer>> generateClusterAssignments(List<Integer> clusterLabels,
+                                                                           DenseVector outlierScoresVector) {
+        // A map value of TreeMap<Double, Integer>> provides a sorting of the entries by outlier score.
+        Map<Integer, TreeMap<Double, Integer>> clusterAssignments = new HashMap<>();
+        for (int i = 0; i < clusterLabels.size(); i++) {
+            Integer clusterLabel = clusterLabels.get(i);
+            Double outlierScore = outlierScoresVector.get(i);
+            if (clusterAssignments.containsKey(clusterLabel)) {
```
Craigacp, line 662: Can this be `computeIfAbsent`?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+                    // score out of the remaining points assigned this cluster.
+                    Entry<Double, Integer> entry = outlierScoreIndexTree.pollFirstEntry();
+                    clusterExemplars.add(new ClusterExemplar(clusterLabel, entry.getKey(), data[entry.getValue()]));
+                }
+            }
+            else {
+                for (int i = 0; i < numExemplarsThisCluster; i++) {
+                    // Note that for outliers the last node is polled from the tree, which has the highest outlier score
+                    // out of the remaining points assigned this cluster.
+                    Entry<Double, Integer> entry = outlierScoreIndexTree.pollFirstEntry();
```
Craigacp, line 705: The comment says this polls the last node, but the code is polling the first node?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+     * A Runnable implementation of the core distance calculation for parallelization.
+     * To be used with an {@link ExecutorService}
+     */
+    private static class CoreDistanceRunnable implements Runnable {
+
+        final private SGDVector[] data;
+        final private int numNeighbors;
+        final private Distance distanceType;
+        final private int point;
+        DenseVector coreDistances;
```
Craigacp, line 786: Why isn't `coreDistances` final?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+
+                // Check at which position in the nearest distances the current distance would fit
+                int neighborIndex = numNeighbors;
+                while (neighborIndex >= 1 && distance < kNNDistances[neighborIndex-1]) {
+                    neighborIndex--;
+                }
+
+                // Shift elements in the array to make room for the current distance
+                if (neighborIndex < numNeighbors) {
+                    for (int shiftIndex = numNeighbors-1; shiftIndex > neighborIndex; shiftIndex--) {
```
Craigacp, line 817: Similar comments to the version above about use of `Arrays.binarySearch` and `System.arraycopy`.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/TrainTest.java
```diff
@@ -0,0 +1,107 @@
+package org.tribuo.clustering.hdbscan;
```
Craigacp, line 1: Copyright & license header.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/package-info.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2015-2021, Oracle and/or its affiliates. All rights reserved.
```
Craigacp, line 2: Copyright year 2021.

Clustering/Hdbscan/src/main/resources/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
Craigacp, line 1: This and the `THIRD_PARTY_LICENSES.txt` files should be symlinks to the ones in the root directory of the repo, rather than copies of those files. That way we don't need to update them everywhere when versions change.

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright (c) 2015-2021, Oracle and/or its affiliates. All rights reserved.
```
Craigacp, line 2: Copyright year 2021.

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
```diff
@@ -0,0 +1,228 @@
...
...,5,5,5,4,4,5,5,4,5,5,4,3,4,3,5,0,3,5,4,5,3,3,5,4,3,5,5,5,4,5,3,5,5,5,4,5,5,0,5,5,5,4,3,4,5,0,5,5,3,5,0,5,5,5,5,5,5,4,3,5,3,4,0,5,5,0,5,5,3,5,5,3,5,5,5,5,4,3,3,5,4,3,3,3,5,5,5,5,4,3,3,5,5,5,3,5,3,3,4,5,5,5,4,5,5,5,5,5,5,5,5,4,5,3,5,5,4,5,5,4,3,5,5,5,5,5,4,5,0,5,3,5,3,3,5,5,4,5,5,5,5,4,5,4,5,5,4,5,5,0,4,5,5,4,4,5,5,5,5,4,5,5,5,3,5,5,5,4,5,4,4,5,4,4,4,5,4,3,3,3,4,3,5,4,3,5,5,3,3,5,3,4,3,5,5,5,5,3,5,5,3,5,4,3,0,5,4,4,3,5,4,4,3,5,4,5,5,5,4,3,4,5,5,4,3,3,5,5,5,5,5,5,3,5,3,5,5,4,4,5,5,3,5,5,5,5,5,5,3,5,3,0,5,3,3,5,5,5,5,3,3,4,3,4,5,5,3,5,4,5,4,3,3,4,5,5,3,5,4,3,5,3,3,4,0,5,5,5,4,4,5,3,3,5,5,5,5,5,5,3,5,5,5,4,4,5,5,5,3,5,5,3,0,4,4,3,4,4,4,5,0,5,3,5,5,4,5,5,3,5,4,3,5,5,5,4,4,5,5,5,5,3,5,4,4,5,4,5,5,5,3,4,3,5,5,4,5,4,4,0,4,4,4,4,4,3,5,3,5,4,5,4,4,3,3,3,5,5,5,5,5,3,5,3,5,4,5,4,4,4,3,3,3,3,5,5,5,4,3,5,3,5,3,4,5,5,5,4,5,3,3,3,4,3,3,5,4,5,3,5,3,5,5,3,5,3,5,3,4,4,4,3,5,5,5,3,4,4,3,5,3,4,4,4,4,5,3,5,5,4,5,5,5,5,3,5,5,5,4,3,4,4,5,4,4,3,5,5,5,5,3,5,5,5,4,4,3,3,3,5,4,3,0,4,4,5,0,5,5,3,3,4,5,4,5,3,3,5,4,5,5,3,5,4,5,5,5,5,4,4,5,5,5,5,5,0,3,4,5,4,4,5,3,4,5,5,5,3,3,3,0,0,3,3,5,3,3,5,5,3,5,5,5,4,5,4,3,4,3,0,3,5,4,4,3,5,4,4,4,3,5,5,3,4,4,3,3,4,5,4,3,3,3,4,5,3,3,4,4,5,3,5,5,3,5,4,4,5,3,3,3,3,4,0,5,4,5,5,4};
```
Craigacp, line 79: Is it worth putting these cluster labels into the csv files and loading them that way?

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
```diff
@@ -0,0 +1,228 @@
...
+
+        assertArrayEquals(expectedLabelPredictions, actualLabelPredictions);
+        assertArrayEquals(expectedOutlierScorePredictions, actualOutlierScorePredictions);
+    }
+
+    @Test
+    public void testBasicTrainPredict() {
+        runBasicTrainPredict(t);
```
Craigacp, line 176: Can you add `Helpers.testModelSerialization(model,ClusterID.class);` to one of these tests. The model serialization helper checks the provenance and model serialization and we want that check on all model implementations.

distribution/pom.xml
```diff
@@ -195,6 +195,11 @@
             <artifactId>tribuo-clustering-kmeans</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.tribuo</groupId>
+            <artifactId>tribuo-clustering-hdbscan</artifactId>
+            <version>4.2.0-SNAPSHOT</version>
```
Craigacp, line 201: Can you replace the version with `${project.version}`?

Craigacp (changes requested):
This looks good. Most of the comments are about code style to fit in with the rest of Tribuo. There's one thing that looks like it might be a bug (or a mismatch between the comment and the code), and then a few things where it looks like it might be using a heavier data structure than necessary.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+                }
+
+                for (int neighbor = 0; neighbor < data.length; neighbor++) {
+                    if (point == neighbor)
+                        continue;
+                    double distance = getDistance(data[point], data[neighbor], distanceType);
+
+                    // Check at which position in the nearest distances the current distance would fit
+                    int neighborIndex = numNeighbors;
+                    while (neighborIndex >= 1 && distance < kNNDistances[neighborIndex - 1]) {
```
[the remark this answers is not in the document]
geoffreydstewart: Typically k, and therefore numNeighbors, are pretty small. Probably average between 5 and 8. Not sure it's worth it, but not opposed to it either.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -0,0 +1,828 @@
...
+                        neighborIndex--;
+                    }
+
+                    // Shift elements in the array to make room for the current distance
+                    if (neighborIndex < numNeighbors) {
+                        for (int shiftIndex = numNeighbors - 1; shiftIndex > neighborIndex; shiftIndex--) {
```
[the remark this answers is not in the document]
geoffreydstewart: Yes, my IDE also complains to me. I tried the resulting arraycopy, and it was rather ugly. I felt this was more readable, and the values are pretty small.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: I wasn't sure that for this case a PriorityQueue<Pair<Integer, List<Integer>>> will bring a significant improvement over the TreeMap. Any link or literature to share?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: I didn't see a nice way to do this with computeIfAbsent. If you have a code snippet, I'll take it though.

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
geoffreydstewart: I did consider this, but thought it might be cumbersome to have a few different csv files to refer to. I preferred having the values right in the test, but am not strongly opposed to using csv files either.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
Craigacp: I think this should be equivalent?

```
--- a/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
+++ b/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
@@ -665,14 +665,8 @@ final public class HdbscanTrainer implements Trainer<ClusterID> {
         for (int i = 0; i < clusterLabels.size(); i++) {
             Integer clusterLabel = clusterLabels.get(i);
             Double outlierScore = outlierScoresVector.get(i);
-            if (clusterAssignments.containsKey(clusterLabel)) {
-                clusterAssignments.get(clusterLabel).put(outlierScore, i);
-            }
-            else {
-                TreeMap<Double, Integer> outlierScoreIndexTree = new TreeMap<>();
-                outlierScoreIndexTree.put(outlierScore, i);
-                clusterAssignments.put(clusterLabel, outlierScoreIndexTree);
-            }
+            Map<Double, Integer> tm = clusterAssignments.computeIfAbsent(clusterLabel, (j) -> new TreeMap<>());
+            tm.put(outlierScore,i);
         }
         return clusterAssignments;
     }
```

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
Craigacp: Ok, that's fine. I agree it makes it easier to use the same test data for different algorithms.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/ExtendedMinimumSpanningTree.java
```diff
@@ -0,0 +1,187 @@
...
+            if (vertexOne != vertexTwo)
+                this.edges[vertexTwo].add(vertexOne);
+        }
+    }
+
+    /**
+     * Quicksorts the graph by edge weight in ascending order.  This quicksort implementation is
+     * iterative and in-place.
+     */
+    void quicksortByEdgeWeight() {
```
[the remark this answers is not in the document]
geoffreydstewart: Good point. I refactored this to use `Arrays.sort()`

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanCluster.java
```diff
@@ -0,0 +1,154 @@
...
+                this.propagatedLowestChildDeathLevel = this.deathLevel;
+            if (this.propagatedLowestChildDeathLevel < this.parent.propagatedLowestChildDeathLevel)
+                this.parent.propagatedLowestChildDeathLevel = this.propagatedLowestChildDeathLevel;
+
+            // If this cluster has no children, it must propagate itself:
+            if (!this.hasChildren) {
+                this.parent.propagatedStability+= this.stability;
+                this.parent.propagatedDescendants.add(this);
+            }
+            else {
```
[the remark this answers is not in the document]
geoffreydstewart: I've done my best to go through all the new code to follow this standard.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Yep, that works great, thanks

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: That is a bug! Thanks for catching this. For harder datasets this subtle issue would really effect predictions.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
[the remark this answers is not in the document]
Craigacp: Ok, in that case I think a comment saying that the expected range of `k` is small and so a binary search is not worthwhile. That way if it does become a bottleneck later with large `k` we'll know what to change. Similarly a comment next to the bit of code that does the manual arraycopy mentioning the small size.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
Craigacp: It looks like `TreeMap` implementation traverses the tree to get the last entry (https://github.com/openjdk/jdk11u/blob/master/src/java.base/share/classes/java/util/TreeMap.java#L2134), so the poll on the priority queue might be better. But it doesn't matter too much.

geoffreydstewart: Thanks for providing this reference (that I should have checked myself) - performance always matters and O(1) is certainly more desirable than O(log n). I tried changing this with a `PriorityQueue<Pair<Integer,List<Integer>>> pq = new PriorityQueue<>(Comparator.comparing(Pair::getA));` but was finding the `contains` method from the `PriorityQueue` a little awkward, to check if the queue contains a `Pair<Integer,List<Integer>>` so that `List<Integer>` can be appended to.  I think we should stick with TreeMap here for now.

Craigacp: Ok, that's fine then.

Craigacp (dismissed):
LGTM, thanks.

Craigacp (approved):

Issue #158, opened 2021-08-11
Tribuo KmeansTrainer doesn't work with Java 11 (or any later version) + SecurityManager because of using ForkJoinPool


spbjss:
**Describe the bug**
From Java 11, the security policies doesn’t get propagated to ForkJoinPool worker threads if a SecurityManager is present ( https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ForkJoinPool.html )
Kmeans depends on ForkJoinPool to run the training tasks: https://github.com/oracle/tribuo/blob/v4.1.0/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java#L197 
So, any java application relay on SecurityManager and security policies won't able to work with Tribuo Kmeans because of AccessControlException.

**To Reproduce**
We have a Java application with security manager and security policy, our Java application could work with Tribuo Kmeans and Java 8. But when the Java application (with Tribuo Kmeans) work with Java 11, Java 14 or Java 15, there would be AccessControlException:
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "modifyThreadGroup")

**Expected behaviour**
The security policy defined by the Java application should be applied to any threads.

**Screenshots**
N/A.

**System information:**

 - OS: [e.g., Linux]
 - Java Version: [e.g.,  11, 14 & 15]
 - JDK Vendor: [e.g., Oracle & OpenJDK ]

**Additional context**
Reference URLs:
https://stackoverflow.com/questions/55333290/jdk11-security-policies-doesn-t-get-propagated-to-forkjoinpool-worker-threads
http://5.9.10.113/66435138/how-to-perform-the-parallel-execution-under-a-collection-when-securitymanager-is
http://mail.openjdk.java.net/pipermail/core-libs-dev/2013-December/024044.html
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ForkJoinPool.html

Craigacp:
Is there a specific security.policy file that triggers this or is it any installation of a SecurityManager?

Craigacp:
Also can you paste the stack trace you get? Tribuo doesn't use the default ForkJoinPool anywhere so I'm surprised we're hitting this issue.

ylwu-amzn:
@Craigacp Thanks a lot! BTW, are you running on Java 8?

Craigacp:
[...]

PR #197, opened 2021-12-02
Single threaded K-Means training no longer uses a ForkJoinPool

Craigacp:
### Description
`KMeansTrainer` uses a `ForkJoinPool` to provide parallel computation of the e and m steps. This FJP is used even when operating in a single thread. The FJP requires extra permissions (namely "modifyThread") when running under a security manager, and in some security managers (notably OpenSearch's SecureSM) the FJP does not work at all due to a permissions issue inside OpenJDK. This means that the FJP breaks single threaded k-means training when running under a security manager, even though the FJP isn't necessary for sequential training. This change moves the FJP so it's only instantiated in multi-threaded execution paths, and also tidies up some of the code in `KMeansTrainer`. When using the FJP in `KMeansTrainer` and `KNNModel` it now uses a custom `ForkJoinWorkerThreadFactory` to allow appropriate security privileges to be granted to the FJP threads, so now parallel computation will work under `SecureSM` when given the appropriate security policy.

There's also a bug fix for the centroid computation when using weighted examples, as previously the weights were required to sum to 1 (but this is not guaranteed by the contract of the weights).

### Motivation
The FJP in `KMeansTrainer` interacts poorly with `SecureSM` used in OpenSearch as the security manager. This change means the FJP isn't used in sequential training allowing those runs to work under a `SecureSM`. With the custom `ForkJoinWorkerThreadFactory` now both `KMeansTrainer` and `KNNModel` should work under `SecureSM` with the appropriate security policy.

This issue was reported in #158.

### DIFF
--- a/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
+++ b/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
@@ -461,13 +461,13 @@ public class KMeansTrainer implements Trainer<ClusterID> {
             DenseVector newCentroid = centroidVectors[e.getKey()];
             newCentroid.fill(0.0);
 
-            int counter = 0;
+            double weightSum = 0.0;
             for (Integer idx : e.getValue()) {
                 newCentroid.intersectAndAddInPlace(data[idx], (double f) -> f * weights[idx]);
-                counter++;
+                weightSum += weights[idx];
             }
-            if (counter > 0) {
-                newCentroid.scaleInPlace(1.0 / counter);
+            if (weightSum != 0.0) {
+                newCentroid.scaleInPlace(1.0 / weightSum);
             }
         };
 
--- a/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
+++ b/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
@@ -46,6 +46,7 @@ import java.util.SplittableRandom;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.IntStream;
@@ -138,8 +139,7 @@ public class KMeansTrainer implements Trainer<ClusterID> {
     /**
      * for olcut.
      */
-    private KMeansTrainer() {
-    }
+    private KMeansTrainer() { }
 
     /**
      * Constructs a K-Means trainer using the supplied parameters and the default random initialisation.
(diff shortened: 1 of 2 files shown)


Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
```diff
@@ -495,4 +505,12 @@ public int compareTo(OutputDoublePair<T> o) {
         }
     }
 
+    /**
+     * Used to allow FJPs to work with OpenSearch's SecureSM.
+     */
+    private static final class CustomForkJoinWorkerThreadFactory implements ForkJoinPool.ForkJoinWorkerThreadFactory {
```
ylwu-amzn, line 511: See same method defined in both `KMeansTrainer` and here, not sure if can define in some common place then these 2 classes can reuse directly. So we can just change one place if we need to tune this method in future.

Craigacp: I agree it's not a good idea to have duplicated code, but I don't want to have the `CustomForkJoinWorkerThreadFactory` be a public class that becomes part of Tribuo's API. As Tribuo isn't using Java modules to guard implementation packages yet I think the options are make the class public or duplicate the code. If we end up requiring this in more than two places then we can spend some time and design a proper entry point (maybe a method on `org.tribuo.util.Util` which constructs a `ForkJoinPool` with the required parallelism).

ylwu-amzn: Yeah, looks a factory/util class generates `ForkJoinPool` is one option, so we can avoid duplicate `CustomForkJoinWorkerThreadFactory` and `System.getSecurityManager() == null ? new ForkJoinPool(numThreads) : new ForkJoinPool(numThreads, THREAD_FACTORY, null, false);`

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
```diff
@@ -63,6 +67,10 @@
  * of threads used in the training step. The thread pool is local to an invocation of train,
  * so there can be multiple concurrent trainings.
  * <p>
+ * Note parallel training uses a {@link ForkJoinPool} which requires that the Tribuo codebase
+ * is given the "modifyThread" and "modifyThreadGroup" privileges when running under a
```
ylwu-amzn, line 71: I think `modifyThreadGroup` is optional for `java.lang.SecurityManager`, only needed as OpenSearch `SecureSM` has stricter check of `modifyThreadGroup`

Craigacp: So I agree, but the `modifyThreadGroup` check is triggered in the codepath, it's just that `java.lang.SecurityManager` doesn't care about it. So if there are other projects which use their own subclass of `SecurityManager` might need to care about it, and explaining that level of complexity didn't seem necessary, especially as the users of Tribuo might not understand the distinction.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
Craigacp: For the moment I'd like to leave it as is. It's all private so we can always change it later to point at a public endpoint when we have more pieces of code which need this functionality.

ylwu-amzn: Sure, sound good to me

pogren (approved):
KMeansTrainer.train 
- conditionally instantiates a FJP if #threads is > 1.  
	- uses a FJP with a custom thread factory which uses the access controller - if a security manager is present
- cluster assignments are performed in a function called eStepFunc which can be used in parallel or in single-threaded mode.  It looks like a straightforward refactoring with no new logic (for the actual work that's being done)
- similar refactor of mStep  

KNNModel only updates the FJP to use the custom factory which uses the access controller.

PR #198, opened 2021-12-02
Adds setInvocationCount to HdbscanTrainer

Craigacp:
### Description
Adds the `setInvocationCount` method to `HdbscanTrainer` and tidies up a couple of small things in the HDBSCAN package.

It also enforces that the default method `Trainer.train(Dataset<T>,Map<String,Provenance>,int)` accepts -1 to mean "do not change the invocation count before training" which is the implied semantics for `INCREMENT_INVOCATION_COUNT`.

### Motivation
Improves integration with the reproducibility framework.

### DIFF
--- a/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
+++ b/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
@@ -27,6 +27,7 @@ import org.tribuo.clustering.ClusterID;
 import org.tribuo.clustering.ImmutableClusteringInfo;
 import org.tribuo.math.la.DenseVector;
 import org.tribuo.math.la.SGDVector;
+import org.tribuo.math.la.SparseVector;
 import org.tribuo.provenance.ModelProvenance;
 import org.tribuo.provenance.TrainerProvenance;
 import org.tribuo.provenance.impl.TrainerProvenanceImpl;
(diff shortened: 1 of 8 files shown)


Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -147,10 +148,14 @@ public HdbscanModel train(Dataset<ClusterID> examples, Map<String, Provenance> r
         }
         ImmutableFeatureMap featureMap = examples.getFeatureIDMap();
 
-        SGDVector[] data = new DenseVector[examples.size()];
+        SGDVector[] data = new SGDVector[examples.size()];
         int n = 0;
         for (Example<ClusterID> example : examples) {
-            data[n] = DenseVector.createDenseVector(example, featureMap, false);
+            if (example.size() == featureMap.size()) {
```
geoffreydstewart, line 154: I believe all the unit tests will currently have `example.size() == featureMap.size()`. Is there any value to add a tests which provides sparse vectors?

geoffreydstewart (commented):
The cleanup related changes look good. Thanks for taking care of those. I'll leave it to the experts to check the changes for setting the invocation count and the new, corresponding test.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
Craigacp: We usually run a sparse and a dense dataset through to make sure nothing breaks, but those tests have been there for a long time and are probably less valuable now as the vector creation infrastructure is more universally used. I probably should add one in this case though as a smoke test to make sure it executes.

Craigacp: I added the small sparse data test that we use in `KMeans`.

pogren (approved):
all changes look straightforward and sensible.

PR #201, opened 2021-12-09
KMeans DenseVector support

Craigacp:
### Description
Relaxed the method signatures in `KMeansTrainer` to allow the use of `DenseVector` or `SparseVector`, and changed the train & predict methods to dynamically pick between sparse and dense vectors. This is similar to the change for #112.

This speeds up k-means in dense spaces as dense vectors are faster than a sparse vector operating in a dense space.

This changes the signature of the protected `mStep` method, relaxing one of the argument types from `SparseVector[]` to `SGDVector[]`. This change means the train method will no longer call subclasses which override the `mStep` method, and so it's a breaking change for subclasses of `KMeansTrainer`. If users have tagged their override with `@Override` then the compiler will warn them and it should be a one line change.

### Motivation
Improves the speed of K-means when working in dense spaces.


JackSullivan (approved):
Based on our conversation I think this is an acceptably marginal change.

PR #202, opened 2021-12-09
Add the HDBSCAN* clustering tutorial, and add a small fix for predictions

geoffreydstewart:
Signed-off-by: Geoffrey Stewart <geoff.stewart@oracle.com>

### Description
The tutorial for HDBSCAN* clustering is being added, and a fix for predictions is included.

### Motivation
The tutorials are the primary way to instruct the community about new features. This is a tutorial for the recently added HDBSCAN* implementation.

A small fix which improves the prediction logic is included here. To summarize, a TreeMap is replaced with a PriorityQueue. There can be rare cases where the objects added to the TreeMap have duplicate keys, which causes the wrong representation of cluster assignments. The rare cases include some very small datasets, and datasets with duplicate, or very near duplicate points.


Craigacp (approved):
LGTM thanks.

PR #203, opened 2021-12-10
Moving ONNX export utils out into a separate module

Craigacp:
### Description
Moves the `org.tribuo.onnx` classes from `tribuo-core` to `tribuo-util-onnx`, which has no dependencies on the rest of Tribuo. Then moved `ONNXExportable` to `org.tribuo`, and moved the model proto construction method to `ONNXExportable` from `ONNXContext`.

### Motivation
Splitting out the core ONNX model writing code into a separate jar will make it possible for the rest of the JVM ecosystem to use it to build out ONNX model support without inducing a Tribuo dependency. This will make it easier if we decide to upstream that support later, and will hopefully encourage contributions to fill out the support with the rest of the ONNX operators.


Core/src/main/java/org/tribuo/ONNXExportable.java
```diff
@@ -47,9 +58,39 @@
...
+    }
+
     /**
      * Exports this {@link org.tribuo.Model} as an ONNX protobuf.
-     * @param domain A reverse-DNS name to namespace the model (e.g., org.tribuo.classification.sgd.linear).
+     *
+     * @param domain       A reverse-DNS name to namespace the model (e.g., org.tribuo.classification.sgd.linear).
      * @param modelVersion A version number for this model.
      * @return The ONNX ModelProto representing this Tribuo Model.
      */
```
JackSullivan, line 96: Given the new `buildModel` method above, shouldn't this be deprecated or removed?

JackSullivan (commented):
A minor question about `ONNXExportable` interface, otherwise looks good.

Core/src/main/java/org/tribuo/ONNXExportable.java
Craigacp: Unfortunately there's no typing relationship between `ONNXExportable` and `Model` so I can't get access to the members necessary to default the method, and I don't really want to make `ONNXExportable` have a bunch of extra methods to expose that information as it makes adding the interface to things like `SequenceModel` tricky.

JackSullivan (approved):
Looks good to me.

PR #214, opened 2022-02-08
These are experimental changes to measure neighbour performance

geoffreydstewart:
This PR is for illustration and feedback only. These changes are not intended to be merged, but they are useful for measuring neighbour query implementations, and can be used again once the k-d tree implementation is complete. Probably I'll just close this PR in a few days.

### DIFF
--- a/Clustering/Core/src/test/java/org/tribuo/neighbours/SGDVectorsFromCSV.java
+++ b/Clustering/Core/src/test/java/org/tribuo/neighbours/SGDVectorsFromCSV.java
@@ -1,0 +1,47 @@
+package org.tribuo.neighbours;
+
+import org.junit.jupiter.api.Test;
+import org.tribuo.math.la.DenseVector;
+import org.tribuo.math.la.SGDVector;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SGDVectorsFromCSV {
+
+    private static SGDVector[] getSGDVectorsFromCSV(String filePath, boolean fileContainsHeader) {
+        List<SGDVector> vectorList = new ArrayList<>();
+        try (BufferedReader br = new BufferedReader(new FileReader(filePath))) {
+            String line;
+
+            if (fileContainsHeader) {
+                line = br.readLine();
+                System.out.println("The header line: " + line + " was ignored.");
+            }
+
+            while ((line = br.readLine()) != null) {
+                String[] values = line.split(",");
+                double[] doubles = new double[values.length];
+                for (int i=0; i<values.length; i++) {
+                    doubles[i] = Double.parseDouble(values[i]);
+                }
+                SGDVector vector = DenseVector.createDenseVector(doubles);
+                vectorList.add(vector);
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+            System.out.println("\nThere is a problem with the file " + filePath);
+        }
+        return vectorList.toArray(new SGDVector[0]);
+    }
+
+    @Test
+    public void testTheMethod() {
+        String filename = "src/test/resources/basic-gaussians.csv";
+        SGDVector[] vectorArr = getSGDVectorsFromCSV(filename, true);
+        System.out.println("An array of " + vectorArr.length + " SGDVectors was read from the file " + filename);
+    }
+}
--- a/Clustering/Core/src/test/java/org/tribuo/neighbours/TestNeighboursPerf.java
+++ b/Clustering/Core/src/test/java/org/tribuo/neighbours/TestNeighboursPerf.java
@@ -1,0 +1,158 @@
+package org.tribuo.neighbours;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.tribuo.DataSource;
+import org.tribuo.Dataset;
+import org.tribuo.Example;
+import org.tribuo.ImmutableFeatureMap;
+import org.tribuo.MutableDataset;
+import org.tribuo.clustering.ClusterID;
+import org.tribuo.clustering.example.GaussianClusterDataSource;
+import org.tribuo.math.distance.DistanceType;
+import org.tribuo.math.la.DenseVector;
+import org.tribuo.math.la.SGDVector;
+import org.tribuo.math.neighbour.NeighboursQuery;
+import org.tribuo.math.neighbour.bruteforce.NeighboursBruteForce;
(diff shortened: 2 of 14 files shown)


Clustering/Core/src/test/java/org/tribuo/neighbours/SGDVectorsFromCSV.java
```diff
@@ -0,0 +1,47 @@
...
+import org.tribuo.math.la.DenseVector;
+import org.tribuo.math.la.SGDVector;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SGDVectorsFromCSV {
```
geoffreydstewart, line 13: Getting SGDVectors from csv's is straight-forward, but using the `GaussianClusterDataSource` seemed easier.

Clustering/Core/src/test/java/org/tribuo/neighbours/TestNeighboursPerf.java
```diff
@@ -0,0 +1,158 @@
...
+import org.tribuo.math.neighbour.bruteforce.NeighboursBruteForceFactory;
+import org.tribuo.math.neighbour.bruteforce.NeighboursBruteForceNaive;
+import org.tribuo.math.neighbour.bruteforce.NeighboursBruteForceNaiveFactory;
+import org.tribuo.util.Util;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+
+public class TestNeighboursPerf {
```
geoffreydstewart, line 26: This test class is in Clustering Core module at this point to leverage `GaussianClusterDataSource`

Clustering/Core/src/test/java/org/tribuo/neighbours/TestNeighboursPerf.java
```diff
@@ -0,0 +1,158 @@
...
+    public void testSingleThreadedQueries() {
+        NeighboursBruteForceFactory nbfFactory = new NeighboursBruteForceFactory(DistanceType.L2, 1);
+        NeighboursBruteForceNaiveFactory nbfnFactory = new NeighboursBruteForceNaiveFactory(DistanceType.L2, 1);
+
+        logger.info("PERFORMING SINGLE THREADED TESTS...");
+        logger.info("");
+        doTestIteration(nbfFactory, nbfnFactory);
+    }
+
+    //@Test
```
geoffreydstewart, line 149: The test annotations are commented here to avoid slowing down the build by default.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -221,64 +224,20 @@ public void setInvocationCount(int newInvocationCount) {
...
-            }
-            executorService.shutdown();
-            try {
-                boolean finished = executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.MINUTES);
-                if (!finished) {
-                    throw new RuntimeException("Parallel execution failed");
-                }
-            } catch (InterruptedException e) {
-                throw new RuntimeException("Parallel execution failed", e);
-            }
```
geoffreydstewart, line 280: The neighbour query functionality added in tribuo-math fits in quite nicely here. The changes in this file will be included as part of a PR that will be raised later, with the integration done for KNN as well. They're just here to illustrate the integration that was used for some initial benchmarks.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -762,6 +724,32 @@ private static double getDistance(SGDVector vector1, SGDVector vector2, Distance
...
 
+    /**
+     * This method performs a mapping from a {@link Distance} to the corresponding {@link DistanceType}.
+     * @deprecated This method is deprecated in version 4.3.
+     *
+     * @param distance The distance metric to employ.
+     * @return A {@link DistanceType} value mapped from the provided parameter.
+     */
+    @Deprecated
+    private static DistanceType getDistanceType(Distance distance) {
```
Craigacp, line 735: I think for the actual integration it would be better to add the `DistanceType` as a field on the `Distance` enum which can be accessed by a getter. That'll remove the extra switch. Then alter the signature of the private methods to accept `DistanceType` rather than `Distance`, and push the conversion up as high as we can (maybe even into postConfig on the trainer, add distance type as a configurable field and throw if both are set).

geoffreydstewart: I believe I understood what suggestions you were making so I've added another commit which pushes the conversion higher. Note the impact it has on the `HdbscanModel`. Again, a separate PR will be raised with these actual integration changes.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
```diff
@@ -91,6 +95,12 @@
 
     @Override
     public Prediction<ClusterID> predict(Example<ClusterID> example) {
+        // A deserialized model which doesn't have distType set needs to handle this case. This check can be removed
+        // in a major version release when the deprecated field no longer needs to be supported.
+        DistanceType distanceType = distType;
```
Craigacp, line 100: I think it would be better to bounce distanceType into distType in the deserialization readObject method. You can see multiple places where we use this trick to upgrade older models (e.g. the regression models to fix the indices bug, or the linear sgd models when they were retrofitted with a superclass).

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
```diff
@@ -91,6 +95,12 @@
...
+        if (distanceType == null) {
+            distanceType = this.distanceType.getDistanceType();
+        }
         SparseVector vector = SparseVector.createSparseVector(example,featureIDMap,false);
```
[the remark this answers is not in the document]
Craigacp: Sounds good.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
Craigacp: Ok, sounds good.

PR #215, opened 2022-02-11
Adds support for merging MeanVarianceAccumulators

Craigacp:
### Description
Adds support for merging two `MeanVarianceAccumulators`. At the moment this class isn't used, but we plan to consolidate all the online mean & variance calculations to use this implementation as there are several implementations in the Tribuo codebase. That will require some complicated serialization management so will happen after we've migrated to protobuf serialization.

Also adds some missing copyright statements for a few files.

### Motivation
Merging two accumulators lets us compute the updates in parallel.

### DIFF
--- a/Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
+++ b/Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


Core/src/main/java/org/tribuo/util/MeanVarianceAccumulator.java
```diff
@@ -187,4 +203,23 @@ public String toString() {
...
+     * @param first The first mean variance accumulator.
+     * @param second The second mean variance accumulator.
+     * @return A merged mean variance accumulator.
+     */
+    public static MeanVarianceAccumulator merge(MeanVarianceAccumulator first, MeanVarianceAccumulator second) {
+        double max = Math.max(first.max, second.max);
+        double min = Math.min(first.min, second.min);
+
+        long count = first.count + second.count;
+        double mean = (first.mean + second.mean) / 2.0;
```
katieyounglove, line 218: This line assumes that the counts of the two accumulators are the same.

Core/src/test/java/org/tribuo/util/MeanVarianceAccumulatorTest.java
```diff
@@ -74,4 +89,36 @@ public void testRandomized() {
...
+        SplittableRandom rng = new SplittableRandom(Trainer.DEFAULT_SEED);
+
+        double[] values = new double[4096];
+
+        for (int i = 0; i < values.length; i++) {
+            values[i] = rng.nextDouble();
+        }
+
+        double[] firstValues = new double[2048];
+        double[] secondValues = new double[2048];
```
katieyounglove, line 103: These two accumulators are the same size which is why we didn't catch the bug in the mean calculation. This test should be revised.

katieyounglove (changes requested):

katieyounglove (approved):
LGTM

PR #216, opened 2022-02-24
These are the changes for the integration with the new DistanceType and NeighbourQuery classes

geoffreydstewart:
This PR contains the changes for the integration for the Hdbscan, KMeans and NearestNeighbour modules with the new DistanceType and NeighbourQuery classes.

There are still some minor changes required in the two Clustering notebooks, to use the new constructors, which will be added once these proposed changes have been initially reviewed.


Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansOptions.java
```diff
@@ -41,10 +41,10 @@
     @Option(longName = "kmeans-num-centroids", usage = "Number of centroids in K-Means. Defaults to 10.")
     public int centroids = 10;
     /**
-     * Distance function in K-Means. Defaults to EUCLIDEAN.
+     * Distance function in K-Means. Defaults to L2 (EUCLIDEAN).
      */
-    @Option(longName = "kmeans-distance", usage = "Distance function in K-Means. Defaults to EUCLIDEAN.")
-    public Distance distance = Distance.EUCLIDEAN;
+    @Option(longName = "distance-type-function", usage = "The type of distance function to use for various distance calculations.")
```
Craigacp, line 46: Can you leave "kmeans" in this argument name? The options classes are designed to be combined (so you could have a single `main` that had both `HdbscanOptions` and `KMeansOptions` in it, so they need names that show which trainer they refer to. I'd missed that the `HdbscanOptions` argument names are not prefixed with `hdbscan` earlier, but they should be.

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
```diff
@@ -95,20 +97,33 @@
...
-        L1
+        L1(DistanceType.L1);
+
+        private final DistanceType distanceType;
+
+        Distance(DistanceType distanceType) {
+            this.distanceType = distanceType;
+        }
+
+        public DistanceType getDistanceType() {
```
Craigacp, line 124: Needs some javadoc on this public method.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNClassifierOptions.java
```diff
@@ -56,10 +57,11 @@
     @Option(longName = "knn-num-threads", usage = "Number of threads to use.")
     public int knnNumThreads = 1;
     /**
-     * Distance metric to use. Defaults to L2.
+     * Distance metric to use. Defaults to L2 (EUCLIDEAN).
      */
-    @Option(longName = "knn-distance", usage = "Distance metric to use.")
-    public Distance knnDistance = Distance.L2;
+    @Option(longName = "distance-type-function", usage = "Distance metric to use.")
```
Craigacp, line 62: Same comment about the argument name as above.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNTrainer.java
```diff
@@ -44,25 +47,42 @@
...
-        COSINE
+        COSINE(DistanceType.COSINE);
+
+        private final DistanceType distanceType;
+
+        Distance(DistanceType distanceType) {
+            this.distanceType = distanceType;
+        }
+
+        public DistanceType getDistanceType() {
```
Craigacp, line 74: Some javadoc for the public method.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNTrainer.java
```diff
@@ -44,25 +47,42 @@
...
+            return distanceType;
+        }
     }
 
-    @Config(mandatory = true, description="The distance function used to measure nearest neighbours.")
+    @Deprecated
+    @Config(mandatory = true, description="The distance function used to measure nearest neighbours. This is now deprecated.")
```
Craigacp, line 80: Both this and `distType` can't be mandatory as it'll crash in postConfig. I think neither of them should be mandatory, like in `KMeansTrainer`.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNTrainer.java
```diff
@@ -85,14 +105,35 @@ private KNNTrainer() {}
...
      * @param distance The distance function.
      * @param numThreads The number of threads to use.
      * @param combiner The combination function to aggregate the k predictions.
      * @param backend The computational backend.
      */
+    @Deprecated
     public KNNTrainer(int k, Distance distance, int numThreads, EnsembleCombiner<T> combiner, Backend backend) {
         this.k = k;
-        this.distance = distance;
+        this.distType = distance.getDistanceType();
```
Craigacp, line 136: Maybe this one and the similar one in `KMeansTrainer` should call `this(k, distance.getDistanceType(), numThreads, combiner, backend)`? That'll remove some redundant code.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
```diff
@@ -319,20 +284,7 @@
...
-            case L2:
-                distanceFunc = (a,b) -> b.l2Distance(a);
-                break;
-            case COSINE:
-                distanceFunc = (a,b) -> b.cosineDistance(a);
-                break;
-            default:
-                throw new IllegalStateException("Unknown distance function " + distance);
-        }
+        BiFunction<SGDVector,SGDVector,Double> distanceFunc = (a,b) -> DistanceType.getDistance(b, a, distType);
```
Craigacp, line 287: Maybe it's better to pass the enum through rather than use a lambda? Alternatively we could push the lambda into the enum itself, and remove the switch in `DistanceType`, so `getDistance` just calls the lambda stored as a field in the enum.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
```diff
@@ -87,22 +92,28 @@
         INNERTHREADPOOL
     }
 
-    private final Pair<SparseVector,T>[] vectors;
+    private final Pair<SGDVector,T>[] vectors;
```
Craigacp, line 95: I'm not sure if this is a serialization safe refactoring. I think it should be, but have you tested loading in a 4.2 KNN into this code? And if you've got such a test, might be worth adding it to the unit tests. We've got a few serialization compatibility tests that have been added when the on disk format has changed slightly.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
```diff
@@ -161,43 +166,24 @@
...
-                    distanceFunc = (a,b) -> b.l2Distance(a);
-                    break;
-                case COSINE:
-                    distanceFunc = (a,b) -> b.cosineDistance(a);
-                    break;
-                default:
-                    throw new IllegalStateException("Unknown distance function " + distance);
-            }
+
+            NeighboursBruteForceFactory nbff = new NeighboursBruteForceFactory(distType, numThreads);
```
Craigacp, line 170: Maybe this should become a field on this object? We can fabricate one during deserialization for old KNNs.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -216,69 +267,25 @@ public void setInvocationCount(int newInvocationCount) {
...
-            executorService.shutdown();
-            try {
-                boolean finished = executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.MINUTES);
-                if (!finished) {
-                    throw new RuntimeException("Parallel execution failed");
-                }
-            } catch (InterruptedException e) {
-                throw new RuntimeException("Parallel execution failed", e);
-            }
+        NeighboursBruteForceFactory nbff = new NeighboursBruteForceFactory(distType, numThreads);
```
Craigacp, line 282: Maybe this should be pushed into a configurable field on this trainer? That way it would be easy to benchmark the different implementations.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -117,28 +135,61 @@ private HdbscanTrainer() {
...
      * @param k The number of nearest-neighbors to use in the initial density approximation.
      * @param numThreads The number of threads.
      */
+    @Deprecated
     public HdbscanTrainer(int minClusterSize, Distance distanceType, int k, int numThreads) {
         this.minClusterSize = minClusterSize;
-        this.distanceType = distanceType;
+        this.distType = distanceType.getDistanceType();
+        this.k = k;
+        this.numThreads = numThreads;
```
Craigacp, line 160: Similar to the other constructors, maybe this should directly call the `HdbscanTrainer(int,DistanceType,int,int)` constructor?

Craigacp (changes requested):

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansOptions.java
geoffreydstewart: Thanks for this explanation, that make sense.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNTrainer.java
geoffreydstewart: Good catch on that one.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
geoffreydstewart: This is a good point, and model serialization should certainly be tested, especially for cases like this. I've added a test which a model generated using the old KNN model. Let me know if you there are some ways the test can be improved - the model filename may be a little verbose, etc.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
geoffreydstewart: I agree that this should be a field. I was thinking it would be better to add these changes when there is more than one `NeighboursQueryFactory` implementation to which it could be assigned. I'm also fine to add these changes now, since the k-d tree should be coming soon. Let me know.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Again, I agree that we'll certainly want the `NeighboursQueryFactory` to be field here, it's just a matter of when to add it.

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
```diff
@@ -215,12 +220,7 @@ public KMeansTrainer(int centroids, int iterations, DistanceType distType, int n
      */
     @Deprecated
     public KMeansTrainer(int centroids, int iterations, Distance distanceType, Initialisation initialisationType, int numThreads, long seed) {
-        this.centroids = centroids;
-        this.iterations = iterations;
-        this.distType = distanceType.getDistanceType();
-        this.initialisationType = initialisationType;
-        this.numThreads = numThreads;
-        this.seed = seed;
+        this(centroids, iterations, distanceType.getDistanceType(), initialisationType, numThreads, seed);
```
Craigacp, line 223: The base constructor already calls postConfig, so this one doesn't need to.

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNTrainer.java
```diff
@@ -132,11 +137,7 @@ public KNNTrainer(int k, DistanceType distType, int numThreads, EnsembleCombiner
      */
     @Deprecated
     public KNNTrainer(int k, Distance distance, int numThreads, EnsembleCombiner<T> combiner, Backend backend) {
-        this.k = k;
-        this.distType = distance.getDistanceType();
-        this.numThreads = numThreads;
-        this.combiner = combiner;
-        this.backend = backend;
+        this(k, distance.getDistanceType(), numThreads, combiner, backend);
```
Craigacp, line 140: Similarly this doesn't need to call postConfig because the base constructor does.

Common/NearestNeighbour/src/test/java/org/tribuo/common/nearest/TestKNN.java
```diff
@@ -184,4 +188,32 @@ public void knnRegressionInvalidExampleTest() {
         });
     }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void deserializeKNNRegressionV42ModelTest() {
+        String serializedModelPath = "src/test/resources/KNNTrainerRegressor_k3_L2_nt2_voting_streams_v4.2.model";
```
Craigacp, line 194: Can you look it up as a classpath resource rather than use an explicit path into the source tree? That makes assumptions about what the working dir is when the tests are executed, and can cause problems when running them in an IDE.

Craigacp (changes requested):
Just a few small fixes.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
Craigacp: I'm fine with making this change later, provided we make it before the next release (which is still many months away).

Common/NearestNeighbour/src/main/java/org/tribuo/common/nearest/KNNModel.java
Craigacp: Later is fine.

Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java
geoffreydstewart: indeed, missed that.

Common/NearestNeighbour/src/test/java/org/tribuo/common/nearest/TestKNN.java
geoffreydstewart: Sure. I occasionally run single tests in Intellij IDEA and didn't face an issue, but this suggestion should provide additional safety.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Yes, the `NeighboursKDTreeFactory` and `NeighboursKDTree` classes which will leverage these discussed changes should be ready well in advance of this timeframe.

Craigacp (approved):
LGTM, thanks.

PR #220, opened 2022-03-04
fix: addressing indeterminate Map orderings in tests

kaiyaok2:
### Description
Using Nondex when running command `mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex` in the `tests` directory after installing all dependencies, several tests in `CSVSaverWithMultiOutputsTest` and `EnsembleExportTest` has flaky results (nondeterministically pass or fail). The reason is that the outputs in the tests are dependent on iterators for hashmaps, which does not guarantee a fixed order by default. (From the Javadoc: "HashMap class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.") For example, in the function `save(Path csvPath, Dataset<T> dataset, Set<String> responseNames)` in `CSVSaver.java`, the code is using the iterator defined for `ImmutableFeatureMap` that deals with a `Collection` of values in a HashMap. Therefore, there is no guaranteed order for the entries in `headerLine[]`, which leads to possible test flakiness.

To resolve the problem, I changed the code to construct a sorted `ArrayList` for all keys in the referenced HashMap to ensure determinate iteration order. I also let `ImmutableLabelInfo.java` and `ImmutableMultiLabelInfo.java` deal with `LinkedHashMap` to ensure deterministic element orders. All tests pass without flakiness after such changes.

### Motivation
The change is necessary as related functionalities / tests may fail when, for example, the Java version upgrades in the future, or when the code is run in a different environment. 

### OCA
https://oca.opensource.oracle.com/api/v1/oca-requests/6710/documents/6730/download

### DIFF
--- a/Classification/Core/src/main/java/org/tribuo/classification/ImmutableLabelInfo.java
+++ b/Classification/Core/src/main/java/org/tribuo/classification/ImmutableLabelInfo.java
@@ -22,7 +22,9 @@ import org.tribuo.ImmutableOutputInfo;
 
 import java.io.IOException;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
@@ -49,21 +51,23 @@ public class ImmutableLabelInfo extends LabelInfo implements ImmutableOutputInfo
 
     private ImmutableLabelInfo(ImmutableLabelInfo info) {
         super(info);
-        idLabelMap = new HashMap<>();
+        idLabelMap = new LinkedHashMap<>();
         idLabelMap.putAll(info.idLabelMap);
-        labelIDMap = new HashMap<>();
+        labelIDMap = new LinkedHashMap<>();
         labelIDMap.putAll(info.labelIDMap);
         domain = Collections.unmodifiableSet(new HashSet<>(labels.values()));
     }
 
     ImmutableLabelInfo(LabelInfo info) {
         super(info);
-        idLabelMap = new HashMap<>();
-        labelIDMap = new HashMap<>();
+        idLabelMap = new LinkedHashMap<>();
+        labelIDMap = new LinkedHashMap<>();
         int counter = 0;
-        for (Map.Entry<String,MutableLong> e : labelCounts.entrySet()) {
-            idLabelMap.put(counter,e.getKey());
-            labelIDMap.put(e.getKey(),counter);
+        List<String> keys = new ArrayList<String>(labelCounts.keySet());
+        Collections.sort(keys);
+        for (String key : keys) {
+            idLabelMap.put(counter,key);
+            labelIDMap.put(key,counter);
             counter++;
         }
         domain = Collections.unmodifiableSet(new HashSet<>(labels.values()));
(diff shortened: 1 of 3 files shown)


kaiyaok2:
Hi there, thanks for the response! 
The `LinkedHashMap` approach is to deal with 2 flaky tests in `EnsembleExportTest.java`: `testHeterogeneousClassificationExport` and `testHeterogeneousMultiLabelExport`. Currently the constructors for `ImmutableMultiLabelInfo` and `ImmutableLabelInfo` are using `Map.Entry<String,MutableLong> e : labelCounts.entrySet()` to setup the 2 HashMaps `idLabelMap` and `LabelIDMap`. We need to make these two maps order deterministic to avoid flakiness in the 2 tests.
Currently nondex is not reporting flakiness in the `tests` directory other than `CSVSaverWithMultiOutputsTest.java` and `EnsembleExportTest.java`, so are the other flaky tests you mentioned in modules other than `tests`?
Also if the `LinkedHashMap` approach is problematic I can definitely first update the PR without including those changes~

Craigacp:
Sure, but I'd prefer the code not to rely upon iteration order in the output infos, as it isn't an invariant guaranteed by the output info contract (and we had a bug in v4.0 and v4.1 due to issues where the regression code was depending on iteration order when it shouldn't have been).  I ran nondex on an internal repo we use for validating Tribuo releases which has all the public Tribuo tests in a single project, and hit the errors you found as well as some others. This branch fixes some of the flakyness, but I've not had chance to run down all of it (https://github.com/Craigacp/tribuo/commit/d9e5825ff450097f84eb7fd88cbc91bdbf63a83f). There were some failures in the way that the single label tests were run, K-Means reproducibility test, and those in the ensemble export you found which haven't been fixed that way, which suggests there is a deeper bug or we're depending on label order when we shouldn't be.

kaiyaok2:
I updated a commit with a fix without switching to `LinkedHashMap`.
I used some breakpoints to find the cause of flakiness in `EnsembleExportTest`. Bascially the bug shall just be label order issues (rather than something deep): 
- the setup of `idLabelMap`(line 63 in `ImmutableMultiLabelInfo.java`) used values retrieved from an iterator from `entrySet()` and then used it as the value corresponding to the integer key `counter`. Then as `entrySet()` allows permutation, `idLabelMap` can be something like [1 -> a, 2 -> b] in one run and [1 -> b, 2 -> a] in another run. 
- in line 177, the test calls `WeightedEnsembleModel.createEnsembleFromExistingModels()`. Then in lines 197 and 203 in `WeightedEnsembleModel.java`, the current code calls the iterator of `ImmutableOutputInfo<T>` to add elements into a list. However, `outputInfo` is an instance of `ImmutableInfoIterator` (which implements the `ImmutableOutputInfo` interface), and its defined iterator is inherited from an iterator of an unordered entryset of a hashmap (line 198 in `ImmutableMultiLabelInfo.java`. Therefore, the iterator does not guarantee any order, so there can be order permutations in the lists `firstList` and `comparisonList`, leading to the exception thrown in line 207 of `WeightedEnsembleModel.java`. This issue can be resolved if we change the lists into sets.
Also I updated the solution to flakiness in `CSVSaver`.

kaiyaok2:
I wonder if I understand your message correctly - is it undesirable to sort the keyset?

kaiyaok2:
I agree that changes to ONNX shall be nontrivial, so I updated this pr to only include changes for `CSVSaver`. Appreciate your time and patience looking into this!

Craigacp (approved):
LGTM thanks.

kaiyaok2:
> Thanks for flagging the iteration order issues. I've got a fix for the ensemble export, and some label ordering flakiness in the other tests which I'll make a PR for shortly.

No problem! Glad it works out

PR #221, opened 2022-03-09
Add the two loops variant of this min heap implementation, which is a slight optimization over the current logic

geoffreydstewart:
Signed-off-by: Geoffrey Stewart <geoff.stewart@oracle.com>

### Description
This is a minor change that splits the existing logic into two loops rather than a single loop containing a conditional instruction.

### Motivation
This change makes a small optimization to the exiting logic. The changes has been extensively tested varying the size of the dataset, and the values for k.


Craigacp (approved):
LGTM

PR #222, opened 2022-03-15
Improve the prediction technique

geoffreydstewart:
Improve the prediction technique by identifying points far away from exemplars as noise, rather than requiring such points to be close to noise point exemplars.

In addition to the unit tests, the quality of these prediction technique changes have been verified using existing notebooks available [in this repo/branch](https://github.com/geoffreydstewart/TribuoHdbscan/tree/post-thesis).


Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
```diff
@@ -115,18 +118,38 @@
...
+                    isNoisePoint = false;
+                }
+                if (distance < minDistance) {
+                    minDistance = distance;
+                    clusterLabel = clusterExemplar.getLabel();
+                    outlierScore = clusterExemplar.getOutlierScore();
+                }
+            }
+            if (isNoisePoint) {
+                outlierScore = noisePointsOutlierScore;
```
Craigacp, line 139: This doesn't update the cluster label to the noise cluster? There's nothing in the if statement on distance to prevent it from setting the cluster, so I'm not sure if it will assign noise clusters at prediction time.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -705,14 +710,15 @@ private static DenseVector calculateOutlierScores(double[] pointNoiseLevels, int
...
+     * Compute the exemplars. These are representative points which are subsets of their clusters, and
      * will be used for prediction on unseen data points.
      *
      * @param data An array of {@link DenseVector} containing the data.
      * @param clusterAssignments A map of the cluster labels, and the points assigned to them.
      * @return A list of {@link ClusterExemplar}s which are used for predictions.
      */
-    private static List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Integer, List<Pair<Double, Integer>>> clusterAssignments) {
+    private static List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Integer, List<Pair<Double, Integer>>> clusterAssignments,
+                                                          DistanceType distType) {
```
Craigacp, line 721: Can you add javadoc for this extra parameter?

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -771,6 +809,7 @@ public TrainerProvenance getProvenance() {
         private final Integer label;
         private final Double outlierScore;
         private final SGDVector features;
+        private Double maxDistToEdge = Double.NEGATIVE_INFINITY;
```
Craigacp, line 812: Can this be final, then modify the get to return Double.NEGATIVE_INFINITY if the field is null? That will make it easier to expose when we make the exemplars public later.

Craigacp (changes requested):
Can you add a test loading in a 4.2 model and asserting that the predictions are the same as they are in 4.2? We've done this before in cases where the on-disk format changed and it's a useful check to make sure nothing broke. Otherwise it looks good.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
geoffreydstewart: Yes, you are right. I've corrected this and added some new assertions to an existing test to ensure noise points are being correctly predicted.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Sure, I've updated the code to construct the `ClusterExemplar`s later on, which enables this variable to be final.

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
```diff
@@ -216,6 +240,52 @@ public void testBasicTrainPredict() {
         runBasicTrainPredict(t);
     }
 
+    @Test
+    public void deserializeHdbscanModelV42Test() {
```
geoffreydstewart, line 244: This test verifies that the predictions are the same as they are in 4.2, and using the debugger I could see that `HdbscanModel`s new `noisePointsOutlierScore` variable is set to 0.0 upon deserialization. Once the cluster exemplars are made public, we could maybe also add some quick assertions that shows calling the `getMaxDistToEdge` method returns `Double.NEGATIVE_INFINITY` to be extra careful here.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -740,29 +743,28 @@ private static DenseVector calculateOutlierScores(double[] pointNoiseLevels, int
...
-                    Entry<Double, Integer> entry = outlierScoreIndexTree.pollFirstEntry();
-                    subsetClusterExemplars.add(new ClusterExemplar(clusterLabel, entry.getKey(), data[entry.getValue()]));
-                }
-
-                // For each of the exemplars in this cluster, iterate the remaining nodes in the tree to find the maximum
-                // distance between the exemplar and the members of the cluster. The other exemplars don't need to be
-                // checked here since they won't be on the fringe of the cluster.
-                for (ClusterExemplar clusterExemplar : subsetClusterExemplars) {
+                // First, get the entries that will be used for cluster exemplars.
+                // Note that for non-outliers, the first node is polled from the tree, which has the lowest outlier
```
Craigacp, line 747: I think this note is wrong now, as there are no outlier exemplars anymore, so it should just be "The first node is polled from the tree...".

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
Craigacp: Sounds good.

Craigacp (changes requested):
Looks good apart from the small note about keeping the comments up to date.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Yes, all the exemplars are now non-outliers, so this was ambiguous. It has been removed.

Craigacp (approved):
LGTM.

PR #225, opened 2022-03-22
Fixes for HashSet iteration order problems

Craigacp:
### Description
When looking at the fix for #220 we found several other issues where Tribuo's behaviour depended on the iteration order of `HashSet`. These fell into a few categories: 

- issues where a test failed due to sorting functions being stable, but the presentation order being slightly non-deterministic (MNB, XGBoost, LinearSGD) - these were left as fixing them required changes in dependencies or intrusive changes to how predictions are generated.
- issues where the output of an evaluation `toString` depended on the label indices - these were fixed by adding/exposing methods to fix the label order, then updating the tests to use those methods.
- `WeightedEnsembleModel` assumed that the output indices of the ensemble members were the same, which in practice only affects ONNX export of ensemble models - this is fixed by tightening up the validation during ensemble creation, and adding ONNX gather ops to ensure that the output indices are consistent across models.
- `CategoricalInfo.uniformSample` and `CategoricalInfo.frequencyBasedSample` used methods which iterate the key or entry set of a `HashMap` to construct the sampling tables. While the outputs were always a valid sample from the distribution, this sample depended on the iteration order of the `HashMap` and so was not necessarily portable between machines or runs of the JVM. This has been fixed by sorting the entries using double's natural ordering before building the tables. This produces different samples to Tribuo 4.2.0 and earlier, but it is now self-consistent and will remain so.

### Motivation
Non-determinism is bad for a reproducible ML library.

### DIFF
--- a/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
+++ b/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015-2022, Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
+++ b/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
@@ -206,6 +206,9 @@ public final class LabelConfusionMatrix implements ConfusionMatrix<Label> {
 
     /**
      * Sets the label order used in {@link #toString}.
+     * <p>
+     * If the label order is a subset of the labels in the domain, only the
+     * labels present in the label order will be displayed.
      *
      * @param newLabelOrder The label order to use.
      */
--- a/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
+++ b/Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
@@ -23,6 +23,7 @@ import org.tribuo.classification.Label;
 import org.tribuo.math.la.DenseMatrix;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
@@ -66,7 +67,8 @@ public final class LabelConfusionMatrix implements ConfusionMatrix<Label> {
     /**
      * Creates a confusion matrix from the supplied predictions, using the label info
      * from the supplied model.
-     * @param model The model to use for the label information.
+     *
+     * @param model       The model to use for the label information.
      * @param predictions The predictions.
      */
     public LabelConfusionMatrix(Model<Label> model, List<Prediction<Label>> predictions) {
@@ -75,9 +77,10 @@ public final class LabelConfusionMatrix implements ConfusionMatrix<Label> {
 
     /**
      * Creates a confusion matrix from the supplied predictions and label info.
-     * @throws IllegalArgumentException If the domain doesn't contain all the predictions.
-     * @param domain The label information.
+     *
+     * @param domain      The label information.
      * @param predictions The predictions.
+     * @throws IllegalArgumentException If the domain doesn't contain all the predictions.
      */
     public LabelConfusionMatrix(ImmutableOutputInfo<Label> domain, List<Prediction<Label>> predictions) {
         this.domain = domain;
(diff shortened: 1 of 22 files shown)


Craigacp:
FYI @kaiyaok2 these are the other issues I found when running down the iteration order determinism. Thanks for bringing that to our attention, in particular the ONNX and `CategoricalInfo` fixes are definite bugs in Tribuo's correctness and reproducibility.

Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
```diff
@@ -199,21 +206,34 @@ private int getIDOrThrow(Label key) {
...
      * Sets the label order used in {@link #toString}.
-     * @param labelOrder The label order to use.
+     *
+     * @param newLabelOrder The label order to use.
      */
-    public void setLabelOrder(List<Label> labelOrder) {
-        this.labelOrder = labelOrder;
+    @Override
+    public void setLabelOrder(List<Label> newLabelOrder) {
+        if (newLabelOrder == null || newLabelOrder.isEmpty()) {
```
JackSullivan, line 214: Do we know what size the labelset should be at this point? Can we easily check for that too while we're at it?

Data/src/main/java/org/tribuo/data/csv/CSVSaver.java
```diff
@@ -33,6 +33,7 @@
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
```
JackSullivan, line 36: Presumably this is an unused import

MultiLabel/Core/src/main/java/org/tribuo/multilabel/ImmutableMultiLabelInfo.java
```diff
@@ -190,6 +190,23 @@ public int hashCode() {
...
     }
 
+    @Override
+    public boolean domainAndIDEquals(ImmutableOutputInfo<MultiLabel> other) {
+        if (size() == other.size()) {
+            for (Map.Entry<Integer,String> e : idLabelMap.entrySet()) {
+                MultiLabel otherLbl = other.getOutput(e.getKey());
+                if (otherLbl == null) {
+                    return false;
+                } else if (!otherLbl.getLabelString().equals(e.getValue())) {
```
JackSullivan, line 200: I don't recall, does `MultiLabel` keep it's label string consistent?

MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
```diff
@@ -162,9 +167,32 @@ public double confusion(MultiLabel predicted, MultiLabel truth) {
...
     }
 
+    /**
+     * Sets the label order used in {@link #toString}.
+     *
+     * @param labelOrder The label order to use.
+     */
+    @Override
+    public void setLabelOrder(List<MultiLabel> labelOrder) {
+        if (labelOrder == null || labelOrder.isEmpty()) {
```
JackSullivan, line 177: Same question as before - can we easily check size here as well?

Regression/Core/src/main/java/org/tribuo/regression/ImmutableRegressionInfo.java
```diff
@@ -334,6 +334,23 @@ public String toReadableString() {
...
     }
 
+    @Override
+    public boolean domainAndIDEquals(ImmutableOutputInfo<Regressor> other) {
+        if (size() == other.size()) {
+            for (Map.Entry<Integer,String> e : idLabelMap.entrySet()) {
+                Regressor otherReg = other.getOutput(e.getKey());
+                if (otherReg == null) {
+                    return false;
+                } else if (!otherReg.getDimensionNamesString().equals(e.getValue())) {
```
JackSullivan, line 344: Same as above - does `Regression` keep dimension ordering consistent?

JackSullivan (commented):
A few very minor points, but otherwise looks good

Classification/Core/src/main/java/org/tribuo/classification/evaluation/LabelConfusionMatrix.java
Craigacp: This actually allows you to reduce the set of labels that you print (and was designed that way many years ago), if you only care about showing a subset of them. However that's not properly documented, nor do I have a test for it, so I should add one.

MultiLabel/Core/src/main/java/org/tribuo/multilabel/ImmutableMultiLabelInfo.java
Craigacp: Yeah, because the domain is actually one `MultiLabel` per dimension and the `MutableMultiLabelInfo.observe` method looks at each label dimension individually and expands the domain that way.

Regression/Core/src/main/java/org/tribuo/regression/ImmutableRegressionInfo.java
Craigacp: Yes, `Regressor` requires that the dimensions are sorted lexicographically. Also it's the same trick as before, the getOutput method returns a single dimension at a time, and they compare by strings.

MultiLabel/Core/src/main/java/org/tribuo/multilabel/evaluation/MultiLabelConfusionMatrix.java
Craigacp: Same answer, it's intentional to allow subsetting, but not documented or tested so I'll do that.

JackSullivan (approved):
Looks good to me

PR #226, opened 2022-03-25
Adds protobuf serialization to FeatureMap, Hasher, and VariableInfo implementations

Craigacp:
### Description
Adds protobuf definitions for `FeatureMap`, `Hasher`, `Transformer`, and `VariableInfo` which are of the form `(int version, String class_name, Any serialized_data)`. Each of the concrete implementations of those classes has its own protobuf which forms the serialized_data field. The higher level objects are redirect hooks for the serialization system, the class name field denotes which class should handle deserialization. Classes which are protobuf serializable implement `org.tribuo.ProtoSerializable` which exposes a single `public T serialize()` method. They must also implement a `public static U deserializeFromProto(int version, String className, Any serializedData)` which rebuilds the object from the protobuf. These deserialize methods are called by `org.tribuo.util.ProtoUtil.Instantiate`, and at the moment `FeatureMap` also has a static `deserialize(FeatureDomainProto proto)` method.

I think there probably need to be more static helpers which manage the serialization and deserialization to input/output streams automatically, and I'm also considering a top level redirection proto which can contain at most one of each serializable type in Tribuo. That will allow us to expose `deserialize` and `serialize` methods on `ProtoUtil` which operate on any Tribuo types, redirecting to the appropriate classes as necessary, and thus get us most of the flexibility back that we lose in a non-`java.io.Serializable` system.

To support the tests I added equals methods to `FeatureMap`, `Hasher` implementations and `VariableInfo` implementations.

There are other protos defined in `tribuo-core.proto` which aren't used in this PR, the design of those is still under consideration (especially the model one, as it's different from the rest), so they could change in future PRs (and as such shouldn't be considered blockers for reviewing this PR).

Note at the moment the `deserializeFromProto` method is accessed reflectively, and tries to change the access permissions on the method so it's accessible. I think that's likely to cause problems with modules in the future, and will require an update to our security policy if we release a version with this in, however I'm currently unsure if we want the deserialization methods to be public or not.

The plan is to roll this out to all Tribuo classes which implement `java.io.Serializable` before the next minor release. Next up after this PR is merged is `Output`, then `Example`, `Prediction` and `Dataset`. Then we'll work through all the models.

This is currently a draft PR as I've removed the generated protos for ease of reviewing. When we're happy with all the code I'll regenerate them and add them to the branch.

### Motivation
`java.io.Serializable` makes it difficult to evolve the in memory representations of objects without drastically affecting the serialization format, plus it's the source of a number of security issues. We've mitigated the security issues by providing a JEP290 filter, however the restraint on evolving classes remains. Adding protobuf serialization should allow us to evolve models past major version boundaries which involve class renaming or replacement (with some caveats).


Craigacp:
@pogren has added an automatic serialization mechanism based on annotations, and I've modified the redirect mechanism a bit after internal discussion. This PR still needs some further cleanup wrt the deserialization methods (as they are a little inconsistent and need to have version number validation everywhere).

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+                        Class<?> fieldClazz = field.getType();
+                        String nameStub = (fieldClazz.isArray() || Collection.class.isAssignableFrom(fieldClazz)) ? "addAll" : "set";
+                        Method setter = findMethod(serializedDataBuilderClass, nameStub, fieldName, 1);
+                        obj = convert(obj);
+                        setter.setAccessible(true);
+                        setter.invoke(serializedDataBuilder, obj);
+                        setter.setAccessible(false);
+                    }
+
+                    ProtoSerializableKeysValuesField pskvf = field.getAnnotation(ProtoSerializableKeysValuesField.class);
```
JackSullivan, line 163: I don't think a field should ever be more than one of `ProtoSerializableField`, `ProtoSerializableKeysValuesField`, or `ProtoSerializableMapValuesField`, and this seems like the place to enforce that.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+        if (ann != null) {
+            return ann.version();
+        } else {
+            return -1;
+        }
+    }
+
+    static <SERIALIZED_CLASS extends Message, PROTO_SERIALIZABLE extends ProtoSerializable<SERIALIZED_CLASS>> Class<SERIALIZED_CLASS> getSerializedClass(PROTO_SERIALIZABLE protoSerializable) {
+        @SuppressWarnings("unchecked")
+        Class<SERIALIZED_CLASS> serializedClass = (Class<SERIALIZED_CLASS>) resolveTypeParameter(ProtoSerializable.class, protoSerializable.getClass(), ProtoSerializable.class.getTypeParameters()[0]);
```
JackSullivan, line 223: If the concrete `protoSerializable` has its own type parameters, does this still work properly?

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+        } else if (obj.getClass().isEnum()) {
+            return ((Enum<?>) obj).name();
+        } else if (obj instanceof int[]) {
+            int[] t = (int[]) obj;
+            return Arrays.stream(t).boxed().collect(Collectors.toList());
+        } else if (obj instanceof long[]) {
+            long[] t = (long[]) obj;
+            return Arrays.stream(t).boxed().collect(Collectors.toList());
+        } else if (obj instanceof float[]) {
+            float[] t = (float[]) obj;
```
JackSullivan, line 255: Is there a reason the logic is different here?

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+            ProtoSerializableKeysValuesField pskvf = field.getAnnotation(ProtoSerializableKeysValuesField.class);
+            ProtoSerializableMapValuesField psmvf = field.getAnnotation(ProtoSerializableMapValuesField.class);
+            if ((psf != null) && (pskvf == null) && (psmvf == null)) {
+                String protoFieldName;
+                if (psf.equals(ProtoSerializableField.DEFAULT_FIELD_NAME)) {
+                    protoFieldName = field.getName();
+                } else {
+                    protoFieldName = psf.name();
+                }
+                if (!fieldNameSet.contains(protoFieldName)) {
```
JackSullivan, line 290: Isn't having a duplicated name here a big problem that shouldn't happen?

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: There's no `FloatStream` so you can't do `Arrays.stream`.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+                floats.add(f);
+            }
+            return floats;
+        } else if (obj instanceof double[]) {
+            double[] t = (double[]) obj;
+            return Arrays.stream(t).boxed().collect(Collectors.toList());
+        } else if (obj instanceof String[]) {
+            return Arrays.asList((String[]) obj);
+        } else {
+            return obj;
```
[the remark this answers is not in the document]
Craigacp: Possibly. This is the minimum set of things that I think should be supported. If there are others then we'll hit them as we roll out protobuf support, and then make decisions on a case by case basis.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: Possibly, though Java allows field shadowing in subclasses. That said it should probably go bang if there is shadowing because it's likely to make the serialized format impossible to reason about.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: Will do.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: I'll check.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
JackSullivan: It looks like this is actually checked it `getFields` further down

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
JackSullivan: We should probably warn the user if we hit this, then.

Core/src/main/java/org/tribuo/hash/HashedFeatureMap.java
```diff
@@ -31,16 +39,49 @@
  * does not contain feature name information, but still works
  * with unhashed features names.
```
JackSullivan, line 40: If it will affect the deserialized object, it should document that the underlying hasher will not deserialize properly.

Core/src/main/resources/protos/tribuo-core.proto
```diff
@@ -0,0 +1,214 @@
...
+option java_package = "org.tribuo.protos.core";
+
+// We use any to encode polymorphism in feature domains, output domains, and model values
+import "google/protobuf/any.proto";
+
+// Import OLCUT's provenance proto
+import "olcut_proto.proto";
+
+/*
+Base model protobuf. Contains the core model fields, along with an Any containing model specific information.
```
JackSullivan, line 34: I don't seen an `Any` here

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
JackSullivan: Looks like this is accounted for in `getFields`

JackSullivan (commented):
Looks good apart from the comments, just a few places where we want more documentation

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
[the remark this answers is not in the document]
Craigacp: I added a check for unexpected types and now it only accepts boxed primitives and Strings in addition to the types we originally checked for.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: Done.

Core/src/main/java/org/tribuo/hash/HashCodeHasher.java
```diff
@@ -16,22 +16,33 @@
...
 
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Hashes names using String.hashCode().
  */
```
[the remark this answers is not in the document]
Craigacp: Done.

Core/src/main/resources/protos/tribuo-core.proto
Craigacp: Fixed.

Core/src/main/java/org/tribuo/hash/HashedFeatureMap.java
Craigacp: Done.

Core/src/main/java/org/tribuo/protos/ProtoUtil.java
Craigacp: I think this is tested in `ProtoUtilTest.testGetSerializedClass()`.

Craigacp:
I fixed the things raised in the review comments, should be good for another review now.

JackSullivan (approved):
Looks good to me

PR #228, opened 2022-03-30
Bumping to liblinear-java 2.44

Craigacp:
### Description
liblinear-java 2.44 has the fix for https://github.com/bwaldvogel/liblinear-java/issues/40, and so this PR migrates Tribuo over to use the new RNG setting methods available in that version. As a consequence it removes all the global locks on liblinear use that were in place from 4.2. This does mean that models trained in 4.2 will not reproduce exactly in 4.3+, though we could consider moving the `forceZero` trick used in the regression tests into the base class and exposing it to the reproducibility framework as a special case.

### Motivation
Allows liblinear models to be trained in parallel with consistent RNG behaviour as compared to the rest of Tribuo.


JackSullivan (approved):
Looks good to me

PR #229, opened 2022-03-30
Exposes views on the HDBSCAN cluster exemplars

### LINKED ISSUE #217: HDBscan clusterExemplars getter
**Is your feature request related to a problem? Please describe.**
When implementing the HDBScan clustering algorithm, there's no way to know the clusters' centroids that the model has generated through its scanning of my dataset. This feature is already been added to the K-Means algorithm, and it would be really useful to also have it in your DB-Scan implementation.

**Describe the solution you'd like**
The `HdbscanModel` class already has a private attribute `clustersExemplars`, which I suppose contain the clusters' centroids. I think just exposing this attribute through a getter method would solve my issue.

Also, it would be nice to get some function to calculate the distance from one of the clusters' centroid to each point contained in said cluster.

Craigacp:
### Description
Adds a get method for cluster exemplars in HDBSCAN which returns copies of the exemplars (as the vector is mutable), along with a method that returns the exemplar in terms of feature names.

### Motivation
Fixes #217.

### DIFF
--- a/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
+++ b/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanModel.java
@@ -19,6 +19,7 @@ package org.tribuo.clustering.hdbscan;
 import com.oracle.labs.mlrg.olcut.util.Pair;
 import org.tribuo.Example;
 import org.tribuo.Excuse;
+import org.tribuo.Feature;
 import org.tribuo.ImmutableFeatureMap;
 import org.tribuo.ImmutableOutputInfo;
 import org.tribuo.Model;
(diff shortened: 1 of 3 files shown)


Craigacp:
@geoffreydstewart would you mind looking over this?

geoffreydstewart (commented):
These changes look good. I'm not sure if it's worth adding some minimal test coverage for the new `getClusters` method. The logic looks quite safe, but it could be instructive for the community. Additionally, there is an existing test `deserializeHdbscanModelV42Test` where we might consider asserting that calling the `getMaxDistToEdge` method on a `ClusterExemplar` from a v4.2 model returns `Double.NEGATIVE_INFINITY`.

Craigacp:
> These changes look good. I'm not sure if it's worth adding some minimal test coverage for the new `getClusters` method. The logic looks quite safe, but it could be instructive for the community. Additionally, there is an existing test `deserializeHdbscanModelV42Test` where we might consider asserting that calling the `getMaxDistToEdge` method on a `ClusterExemplar` from a v4.2 model returns `Double.NEGATIVE_INFINITY`.

I've expanded the tests to cover those points.

JackSullivan (approved):
Looks good to me

PR #243, opened 2022-06-19
Add ModelCard infrastructure (including implementation and testing)

rmahinpei:
Signed-off-by: Romina Mahinpei <mahinpei@student.ubc.ca>

### Description & Motivation
* Implementing & testing of ModelCard
* A new feature enabling users to build an object that contains info about their trained model (info obtained via the model's built-in provenance & user-input)
* Aim of feature is to allow more transparent model reporting 

### Paper reference
[Model Cards for Model Reporting](https://arxiv.org/pdf/1810.03993.pdf)


Interop/ModelCard/src/main/java/TrainingDetails.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public ObjectNode toJson() {
+        ObjectNode datasetDetailsObject = mapper.createObjectNode();
+        datasetDetailsObject.put("schema-version", schemaVersion);
+        datasetDetailsObject.put("training-time", trainingTime);
+        datasetDetailsObject.put("training-set-size", trainingSetSize);
+
+        datasetDetailsObject.put("num-features", numFeatures);
+        ArrayNode featuresArr = mapper.createArrayNode();
+        for (String s : features) featuresArr.add(s);
```
Craigacp, line 105: No one line for loops please, I misread this as iterating the `datasetDetailsObject` line below.

Interop/ModelCard/src/main/java/TrainingDetails.java
```diff
@@ -0,0 +1,121 @@
...
+    public int getTrainingSetSize() {
+        return trainingSetSize;
+    }
+
+    public int getNumFeatures() {
+        return numFeatures;
+    }
+
+    public List<String> getFeatures() {
+        return features;
```
Craigacp, line 86: Similar to the other places, this either should emit an unmodifiable list, or the field itself should be an unmodifiable list. I think in this case it should be the latter.

Interop/ModelCard/src/main/java/TrainingDetails.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public int getNumOutputs() {
+        return numOutputs;
+    }
+
+    public Map<String, Long> getOutputsDistribution() {
+        return outputsDistribution;
```
Craigacp, line 94: Same comment as for `getFeatures` wrt unmodifiability.

Interop/ModelCard/src/main/java/ModelCard.java
```diff
@@ -0,0 +1,74 @@
...
+import org.tribuo.Model;
+import org.tribuo.evaluation.Evaluation;
+import org.tribuo.interop.ExternalModel;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+
+public class ModelCard {
+    private static final ObjectMapper mapper = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
```
Craigacp, line 14: Maybe this could be package private and then all the classes can share the same instance?

Interop/ModelCard/src/main/java/ModelCard.java
```diff
@@ -0,0 +1,74 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.tribuo.Model;
+import org.tribuo.evaluation.Evaluation;
+import org.tribuo.interop.ExternalModel;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+
+public class ModelCard {
```
Craigacp, line 13: For the time being, let's make all the classes in here `final`. I don't want to have to think about subclasses when we evolve this to have feature metrics or correlations in it.

Interop/ModelCard/src/main/java/UsageDetails.java
```diff
@@ -0,0 +1,355 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
+public class UsageDetails implements CommandGroup {
+    private static final ObjectMapper mapper = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);;
+    private static final Logger logger = Logger.getLogger(ModelExplorer.class.getName());
```
Craigacp, line 20: This should be `UsageDetails.class.getName`.

Interop/ModelCard/src/main/java/UsageDetails.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    public ObjectNode toJson() {
+        ObjectNode usageDetailsObject = mapper.createObjectNode();
+        usageDetailsObject.put("schema-version", schemaVersion);
+        usageDetailsObject.put("intended-use", intendedUse);
+        usageDetailsObject.put("intended-users", intendedUsers);
+
+        ArrayNode usesArr = mapper.createArrayNode();
+        for (String s : outOfScopeUses) usesArr.add(s);
```
Craigacp, line 121: No one line for loops.

Interop/ModelCard/src/main/java/UsageDetails.java
```diff
@@ -0,0 +1,355 @@
...
+    )
+    public String modelLicense(CommandInterpreter ci, String license) {
+        modelLicense = license;
+        return("Recorded model license as " + modelLicense + ".");
+    }
+
+    @Command(
+            usage = "<String> Saves UsageDetails to provided destination file and closes CLI."
+    )
+    public String saveUsageDetails(CommandInterpreter ci, String destinationFile) throws IOException {
```
Craigacp, line 330: The command should accept `File` as a type rather than wrapping a String.

Interop/ModelCard/src/main/java/ModelCard.java
```diff
@@ -0,0 +1,74 @@
...
+        return modelCardObject;
+    }
+
+    public void saveToFile(String destinationFile) throws IOException {
+        ObjectNode modelCardObject = toJson();
+        mapper.writeValue(new File(destinationFile), modelCardObject);
+    }
+
+    @Override
+    public String toString() {
```
Craigacp, line 71: These classes should all implement `equals` (and consequently `hashCode`) too. You should be able to autogenerate them in IDEs like IntelliJ.

Interop/ModelCard/src/test/java/EnsembleModelsTest.java
```diff
@@ -0,0 +1,236 @@
...
+        BaggingTrainer<Label> t = new BaggingTrainer<>(lr, VOTING,5);
+        WeightedEnsembleModel<Label> ensemble = (WeightedEnsembleModel<Label>) t.train(train);
+
+        var evaluator = new LabelEvaluator();
+        var evaluation = evaluator.evaluate(ensemble, test);
+
+        // create Model Card object and write to file
+        ModelCard modelCard = new ModelCard(ensemble, evaluation);
+        modelCard.addMetric("overall-accuracy", evaluation.accuracy());
+        modelCard.saveToFile("src/test/output-json/homogeneousClassificationModelCard.json");
```
Craigacp, line 84: This should use `File.createTempFile` and then set the file to `deleteOnExit`.

Interop/ModelCard/src/test/java/EnsembleModelsTest.java
```diff
@@ -0,0 +1,236 @@
...
+        LibSVMClassificationModel svmModel = (LibSVMClassificationModel) svmT.train(train);
+        WeightedEnsembleModel<Label> ensemble = WeightedEnsembleModel.createEnsembleFromExistingModels("Bag+FM", Arrays.asList(bagModel,fmModel,svmModel), FULL_VOTING, new float[]{0.3f,0.5f,0.2f});
+
+        var evaluator = new LabelEvaluator();
+        var evaluation = evaluator.evaluate(ensemble, test);
+
+        // create Model Card object and write to file
+        ModelCard modelCard = new ModelCard(ensemble, evaluation);
+        modelCard.addMetric("overall-accuracy", evaluation.accuracy());
+        modelCard.saveToFile("src/test/output-json/heterogeneousClassificationModelCard.json");
```
Craigacp, line 114: Same comment about using a temp file.

Interop/ModelCard/src/test/java/ExternalModelsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class ExternalModelsTest {
+    @Test
+    public void testExternalModelException() throws IOException {
+        var labelFactory = new LabelFactory();
+        var mnistTrainSource = new IDXDataSource<>(Paths.get("src/test/input-data/train-images-idx3-ubyte.gz"),Paths.get("src/test/input-data/train-labels-idx1-ubyte.gz"),labelFactory);
```
Craigacp, line 21: We can't ship MNIST, so instead just fake up some dataset that has 784 features and 10 labels. As the main result of this test is to throw an exception it doesn't matter if the evaluation is nonsensical.

Interop/ModelCard/src/test/java/NativeModelsTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.nio.file.Paths;
+
+public class NativeModelsTest {
+    @Test
+    public void testClassificationModelCard() throws IOException {
+        var factory = new LabelFactory();
+        var csvLoader = new CSVLoader<>(factory);
+
+        var irisHeaders = new String[]{"sepalLength", "sepalWidth", "petalLength", "petalWidth", "species"};
+        var irisSource = csvLoader.loadDataSource(Paths.get("src/test/input-data/bezdekIris.data"),"species", irisHeaders);
```
Craigacp, line 44: We can't ship irises either, so use some generated data, or randomly generate a csv file and use that.

Interop/ModelCard/src/test/java/NativeModelsTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        // read file and create Model Card object
+        ModelCard modelCardCopy = new ModelCard("src/test/output-json/classificationModelCard.json");
+        assert(modelCard.toString().equals(modelCardCopy.toString()));
+    }
+
+    @Test
+    public void testMultiLabelClassificationModelCard() throws IOException {
+        var factory = new MultiLabelFactory();
+        var trainSource = new LibSVMDataSource<>(Paths.get(".","src/test/input-data/yeast_train.svm"),factory);
```
Craigacp, line 69: Ditto for yeast.

Interop/ModelCard/src/test/java/NativeModelsTest.java
```diff
@@ -0,0 +1,162 @@
...
+        var model = trainer.train(trainData);
+
+        var evaluator = new AnomalyEvaluator();
+        var evaluation = evaluator.evaluate(model,evalData);
+
+        // create Model Card object and write to file
+        ModelCard modelCard = new ModelCard(model, evaluation);
+        modelCard.addMetric("overall-precision", evaluation.getPrecision());
+        modelCard.addMetric("overall-recall", evaluation.getRecall());
+        modelCard.saveToFile("src/test/output-json/anomalyModelCard.json");
```
Craigacp, line 156: Same comment about using `File.createTempFile`.

Interop/pom.xml
```diff
@@ -31,6 +31,7 @@
         <module>OCI</module>
         <module>ONNX</module>
         <module>Tensorflow</module>
+        <module>ModelCard</module>
```
Craigacp, line 34: This should be removed as it's auto-enabled by the profile below.

Interop/ModelCard/src/test/java/EnsembleModelsTest.java
```diff
@@ -0,0 +1,236 @@
...
+        var evaluation = evaluator.evaluate(ensemble, test);
+
+        // create Model Card object and write to file
+        ModelCard modelCard = new ModelCard(ensemble, evaluation);
+        modelCard.addMetric("overall-accuracy", evaluation.accuracy());
+        modelCard.saveToFile("src/test/output-json/homogeneousClassificationModelCard.json");
+
+        // read file and create Model Card object
+        ModelCard modelCardCopy = new ModelCard("src/test/output-json/homogeneousClassificationModelCard.json");
+        assert(modelCard.toString().equals(modelCardCopy.toString()));
```
Craigacp, line 88: You should use `Assertions.assertEquals` to pull in the right thing from JUnit. I assume that the Java assert statement will have a similar effect, but JUnit is what we use everywhere else.

Craigacp (changes requested):
Overall it looks good, but there's a number of things we want to sort out.

First, all the new files need a license header, similar to the ones in other Tribuo source files.

Second, the output-json folder shouldn't be checked in, as far as I can tell it isn't used to validate output. 

Third any data files that are necessary should be in src/test/resources rather than src/test/input-data. That way Maven will know to put them in the classpath of the test classes. We also can't distribute real data files, so let's come up with alternatives where you've used them.

Fourth, the formatting should be consistent, so that's using curly braces on single statements and no one line ifs or fors.

I've made more specific comments on other issues, so we can discuss them in those threads.

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public ModelDetails(Model<?> model) {
+        schemaVersion = "1.0";
+        modelType = model.getClass().getSimpleName();
+        modelPackage = model.getClass().getTypeName();
+        tribuoVersion = model.getProvenance().getTribuoVersion();
+        javaVersion = model.getProvenance().getJavaVersion();
+
+        Map<String,?> parameters = model.getProvenance().getTrainerProvenance().getConfiguredParameters();
+        for (String key : parameters.keySet())
```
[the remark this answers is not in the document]
rmahinpei: For now, I chose to not flatten the map (so it still contains the Provenance objects). When creating the Json version of the ModelCard, everything is processed and flattened properly, so I wasn't quite sure what the advantages of flattening in the constructor would be?

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+    public String getTribuoVersion() {
+        return tribuoVersion;
+    }
+
+    public String getJavaVersion() {
+        return javaVersion;
+    }
+
+    public Map<String, Object> getConfiguredParams() {
+        return configuredParams;
```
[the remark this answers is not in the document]
rmahinpei: For now, I did not wrap the field in `Collections.unmodifiableMap` in the constructor but instead return an unmodifiableMap view to the user in the getter. Not sure if one of these ways is more preferred over the other.

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final ObjectMapper mapper = new ObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);;
+    private final String schemaVersion;
+    private final String modelType;
+    private final String modelPackage;
+    private final String tribuoVersion;
+    private final String javaVersion;
+    private final Map<String, Object> configuredParams = new HashMap<>();
+
+    public ModelDetails(Model<?> model) {
+        schemaVersion = "1.0";
```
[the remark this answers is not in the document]
rmahinpei: I've made it into a public static String for now and can later change it to an enum if it ever becomes preferred.

Interop/ModelCard/src/main/java/ModelDetails.java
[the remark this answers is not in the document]
Craigacp: If you wrap it in the constructor then it means it's unmodifiable elsewhere in the class, and also it doesn't allocate an extra object each time you call the get method. We do wrap in the get method in a few places, but that's usually when the underlying object is mutable in some way, or used to be in the past (so we need to defensively wrap it as old objects will still be mutable).

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+        modelDetailsObject.put("model-type", modelType);
+        modelDetailsObject.put("model-package", modelPackage);
+        modelDetailsObject.put("tribuo-version", tribuoVersion);
+        modelDetailsObject.put("java-version", javaVersion);
+        ObjectNode paramsArr = paramsToJson(null, configuredParams);
+        modelDetailsObject.set("configured-parameters", paramsArr);
+        return modelDetailsObject;
+    }
+
+    private ObjectNode paramsToJson(String name, Map<?,?> params) {
```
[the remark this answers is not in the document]
rmahinpei: Got it - I've now introduced a new field `JsonNode configuredParams` so that this process can instead happen in the constructor. Because of this, I also no longer saw the need to have a `Map configuredParams` (since they would both essentially contain the same info). So, I've removed `Map configuredParams` for now, but if there is a reason for keeping it, I can always add it back.

Interop/ModelCard/src/main/java/ModelDetails.java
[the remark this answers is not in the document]
rmahinpei: I've replaced `Map configuredParams` with `JsonNode configuredParams` for now, so this issue is resolved.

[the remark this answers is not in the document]
rmahinpei: I've replaced Map configuredParams with JsonNode configuredParams (mentioned in thread below) for now, so this issue is resolved (but if we add back Map configuredParams, I'll wrap it in the constructor).

Interop/ModelCard/src/main/java/ModelCard.java
rmahinpei: Done. Right now, for the implementations of `equals` in `ModelDetails`, I'm comparing the text value of the `JsonNode configuredParams` fields for equality rather than directly comparing the objects. This is because in my first constructor for `ModelDetails`, `configuredParams` has an actual type of `ObjectNode` whereas in the second constructor (when building a ModelCard object from a file), the actual type is `JsonNode`. I am aware that the two are used rather interchangeably in practice and in the context that I'm using them in, but I wanted to run this past someone to make sure that it would be okay to do this and just compare their text values in my implementation of `equals`.

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+        modelPackage = model.getClass().getTypeName();
+        tribuoVersion = model.getProvenance().getTribuoVersion();
+        javaVersion = model.getProvenance().getJavaVersion();
+
+        Map<String,?> parameters = model.getProvenance().getTrainerProvenance().getConfiguredParameters();
+        for (String key : parameters.keySet())
+            configuredParams.put(key, parameters.get(key));
+    }
+
+    public ModelDetails(JsonNode modelDetailsJson) throws JsonProcessingException {
```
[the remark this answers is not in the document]
rmahinpei: My reasoning was that because a user has the capability to make a specific chunk of a ModelCard (e.g. just the UsageDetails) and save that to file, they might then later need to also deserialize it too. I've kept the constructor public for now, but if you still think it's best to have it be package private, I'm more than happy to change it.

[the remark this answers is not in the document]
rmahinpei: My reasoning was that because a user has the capability to make a specific chunk of a ModelCard (e.g. just the ModelDetails) and save that to file, they might then later need to also deserialize it too. I've kept the constructor public for now, but if you still think it's best to have it be package private, I'm more than happy to change it.

Interop/ModelCard/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.tribuo</groupId>
+            <artifactId>tribuo-regression-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>2.13.3</version>
```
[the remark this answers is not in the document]
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/ModelCard.java
```diff
@@ -0,0 +1,74 @@
...
+    public ObjectNode toJson() {
+        ObjectNode modelCardObject = mapper.createObjectNode();
+        modelCardObject.set("ModelDetails", modelDetails.toJson());
+        modelCardObject.set("TrainingDetails", trainingDetails.toJson());
+        modelCardObject.set("TestingDetails", testingDetails.toJson());
+        modelCardObject.set("UsageDetails", usageDetails.toJson());
+        return modelCardObject;
+    }
+
+    public void saveToFile(String destinationFile) throws IOException {
```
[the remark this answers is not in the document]
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+        modelDetailsObject.set("configured-parameters", paramsArr);
+        return modelDetailsObject;
+    }
+
+    private ObjectNode paramsToJson(String name, Map<?,?> params) {
+        ObjectNode paramsArr = mapper.createObjectNode();
+        if (name != null) paramsArr.put("className", name);
+        for (var key : params.keySet()) {
+            if (params.get(key) instanceof ConfiguredObjectProvenance) {
+                ConfiguredObjectProvenance prov = (ConfiguredObjectProvenance) params.get(key);
```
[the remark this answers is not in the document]
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/ModelDetails.java
```diff
@@ -0,0 +1,115 @@
...
+                ObjectNode nestedParam = paramsToJson(prov.getClassName(), prov.getConfiguredParameters());
+                paramsArr.set(key.toString(), nestedParam);
+            } else if (params.get(key) instanceof Map) {
+                Map<?,?> map = (Map<?,?>) params.get(key);
+                ObjectNode nestedParam = paramsToJson(null, map);
+                paramsArr.set(key.toString(), nestedParam);
+            } else if (isNumeric(params.get(key).toString())) {
```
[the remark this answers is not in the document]
rmahinpei: Done as suggested by second alternative.

Interop/ModelCard/src/main/java/TestingDetails.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public TestingDetails(Evaluation<?> evaluation) {
+        schemaVersion = "1.0";
+        testingSetSize = evaluation.getPredictions().size();
+    }
+
+    public TestingDetails(JsonNode testingDetailsJson) throws JsonProcessingException {
+        schemaVersion = testingDetailsJson.get("schema-version").textValue();
+        testingSetSize = testingDetailsJson.get("testing-set-size").intValue();
+        Map<String, Double> parsed = mapper.readValue(testingDetailsJson.get("metrics").toString(), Map.class);
```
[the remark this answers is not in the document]
rmahinpei: Yes, but I refactored the code to avoid this.

Interop/ModelCard/src/main/java/ModelCard.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public ModelCard(Model<?> model, Evaluation<?> evaluation) {
+        if (model instanceof ExternalModel) throw new IllegalArgumentException();
+        modelDetails = new ModelDetails(model);
+        trainingDetails = new TrainingDetails(model);
+        testingDetails = new TestingDetails(evaluation);
+        usageDetails = new UsageDetails();
+    }
+
+    public ModelCard(String sourceFile) throws IOException {
```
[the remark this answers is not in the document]
rmahinpei: Done - added both suggested methods.

Interop/ModelCard/src/main/java/TrainingDetails.java
rmahinpei: For now, I have the method emit an unmodifiable list but can always change this to the latter option if it becomes preferred.

Interop/ModelCard/src/main/java/TrainingDetails.java
rmahinpei: For now, I have the method emit an unmodifiable map

Interop/ModelCard/src/main/java/ModelCard.java
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/ModelCard.java
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/UsageDetails.java
```diff
@@ -0,0 +1,355 @@
...
+    public String addPreProcessingStep(CommandInterpreter ci, String step) {
+        preProcessingSteps.add(step);
+        return("Added pre-processing step to list of steps.");
+    }
+
+    @Command(
+            usage = "<int> Remove pro-processing step at specified index (1-indexed)."
+    )
+    public String removePreProcessingStep(CommandInterpreter ci, int index) {
+        preProcessingSteps.remove(index-1);
```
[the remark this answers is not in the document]
rmahinpei: Done as suggested.

Interop/ModelCard/src/main/java/UsageDetails.java
rmahinpei: Done as suggested.

Interop/ModelCard/src/test/java/EnsembleModelsTest.java
rmahinpei: Done as suggested.

Interop/ModelCard/src/test/java/ExternalModelsTest.java
rmahinpei: I've now created a my own mock data as suggested.

Interop/ModelCard/src/test/java/NativeModelsTest.java
rmahinpei: I've now created a my own mock data as suggested.

Interop/ModelCard/src/test/java/NativeModelsTest.java
rmahinpei: I've now created a my own mock data as suggested.

Interop/ModelCard/src/test/java/NativeModelsTest.java
```diff
@@ -0,0 +1,162 @@
...
+        ModelCard modelCardCopy = new ModelCard("src/test/output-json/multiClassificationModelCard.json");
+        assert(modelCard.toString().equals(modelCardCopy.toString()));
+    }
+
+    @Test
+    public void testRegressionModelCard() throws IOException {
+        var factory = new RegressionFactory();
+        var csvLoader = new CSVLoader<>(';', factory);
+
+        var wineSource = csvLoader.loadDataSource(Paths.get("src/test/input-data/winequality-red.csv"),"quality");
```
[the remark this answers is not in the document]
rmahinpei: I've now created a my own mock data as suggested.

Interop/ModelCard/src/test/java/EnsembleModelsTest.java
rmahinpei: Done as suggested.

rmahinpei:
I made edits for most of the suggested changes - I just had a few specific questions about some of the comments (which I've still left as unresolved above). 

I was also curious about the PR failures. I'm still not entirely sure what's causing them (when I run `mvn clean package` locally, the build is always successful). Is there anything I can look into to get that fixed?

Another thing that I was recently thinking about is the precondition I currently have for UsageDetails component of ModelCard. In my head, I always assumed that a user would first build their model & the ModelCard, then save the ModelCard to file, and then afterwards run the CLI to update the UsageDetails component. As a result, my implementation for `saveUsageDetails(CommandInterpreter ci, File destinationFile)` assumes that destinationFile corresponds to a Json representation of a ModelCard and attempts to look for the other components that a ModelCard has before appending UsageDetails. I was wondering whether I should also support the scenario where a user first saves the UsageDetails (i.e. to an empty file) before making their model and ModelCard? I'd have to change parts of the code for when a user saves the ModelCard to file for the first time, but it's definitely doable.

Thanks for all the help!

rmahinpei:
Yes, that makes sense - I have right now combined too many things in `UsageDetails` and splitting would be better.

I can have `UsageDetails` be more like the other detail classes where the user can programatically add some usage-related fields that they would roughly know even before training their model (e.g. intended-use, intended-users, primary-contact) and have `ModelCardCLI` for including fields that would be better known at the very end of the process (e.g. out-of-scope-uses, considerations-list, relevant-factors-list, resources-list, model-citation, model-license).

If I am to follow this design, would it make more sense for ModelCardCLI to just append the extra details to the `UsageDetails` section of a given ModelCard file or would it better to append them in a completely separate section? I personally think that having a separate section might be better (one section would more align with pre-build usage detail and the other with post-build usage details), but I can also see benefits to the other approach as well. I'd appreciate any input on this, and once we're confirmed on the design, I can go ahead and make the changes.

rmahinpei:
I see. Just to confirm: our hope is to offer the user with two options regarding how they can add `UsageDetails` to their ModelCard. They can do it 1) programatically or 2) using the `ModelCardCLI`. 

For first approach, they would use the setters within `UsageDetailsBuilder` to specify any known fields and then build an instance of `UsageDetails`. They can append that instance to their ModelCard object. For the second approach, they would use the `ModelCardCLI`, which will write to `UsageDetailsBuilder`. When the ModelCard is saved, the `UsageDetails` is built and appended to the user's ModelCard. 

My apologies if I'm not fully understanding your vision, so please feel free to correct me. I do also have a follow-up if I roughly explained what the goal is correctly: if we are providing these two approaches, should we support the scenario where a user chooses to specify some fields programatically (approach 1) and then set some other fields using the CLI (approach 2)? Or would we want to throw an error so that the user can only rely on one of these approaches (rather than a combination of both) in order to append `UsageDetails` to their ModelCard?

Interop/ModelCard/src/main/java/ModelDetails.java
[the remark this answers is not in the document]
rmahinpei: Done as suggested.

rmahinpei:
Got it - that makes sense. I'll start separating UsageDetails and ModelCardCLI (my plan is to have it done by the end of Fri, Jul 1). Then, I'll work around `addMetric(...)` in `TestingDetails` so that the class can be immutable (shouldn't take me too long either). Other than that, I know you also mentioned turning all the classes within ModelCard into records, and I was wondering whether this is something I should go ahead and do?

Craigacp:
Moving things to records is lower down the priority list than the other items as I'm still thinking about if we want to depend on the model card code elsewhere in Tribuo (e.g. in ONNX export) in which case we might have to move it down to Java 8. Or decide to punt on the integration into the rest of Tribuo till the next major version when we bump everything to 17. The current set of Java 17 features this PR uses are easy to translate into Java 8, but records would be a bigger change.

Interop/ModelCard/src/main/java/org/tribuo/interop/modelcard/UsageDetailsBuilder.java
```diff
@@ -0,0 +1,146 @@
+package org.tribuo.interop.modelcard;
```
Craigacp, line 1: This file needs the license & copyright header.

Interop/ModelCard/src/main/java/org/tribuo/interop/modelcard/UsageDetailsBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public class UsageDetailsBuilder {
```
Craigacp, line 8: Let's make the class `final` too like the rest of them.

Interop/ModelCard/src/main/java/org/tribuo/interop/modelcard/UsageDetailsBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+    private final List<String> outOfScopeUses = new ArrayList<>();
+    private final List<String> preProcessingSteps = new ArrayList<>();
+    private final List<String> considerations = new ArrayList<>();
+    private final List<String> factors = new ArrayList<>();
+    private final List<String> resources = new ArrayList<>();
+    private String primaryContact = "";
+    private String modelCitation = "";
+    private String modelLicense = "";
+
+    public String getIntendedUse() {
```
Craigacp, line 20: I think it's better to explicitly add the no-args constructor rather than have `javac` silently add it for you, as otherwise it gets quite confusing when you add another constructor, the generated no-args one disappears and things don't compile.

Interop/ModelCard/src/main/java/org/tribuo/interop/modelcard/ModelCardCLI.java
```diff
@@ -0,0 +1,252 @@
...
+    )
+    public String modelLicense(CommandInterpreter ci, String license) {
+        builder.modelLicense(license);
+        return("Recorded model license as " + license + ".");
+    }
+
+    @Command(
+            usage = "<filename> Saves UsageDetails to an existing ModelCard file and close CLI."
+    )
+    public String saveUsageDetails(CommandInterpreter ci, File destinationFile) throws IOException {
```
Craigacp, line 228: Most of the other CLIs we have let you load in different files (e.g. models, datasets) to explore them without restarting the JVM to get a fresh shell. It might be nicer to have this CLI take a filename as an argument for a pre-existing model card, have it load that in on startup, and then you can display it in the CLI while you write the usage details. This method could then save out to a different filename and clear out the usage builder (by making  a fresh one or adding a clear method), clear all the lists, and then the user can load in another model card to write the next set of usage details.

rmahinpei: Done. The user now has the command to start a newUsageDetails, which will effectively allow them to write the UsageDetails for another model card.

Craigacp (approved):
Thanks Romina, this looks great.

PR #245, opened 2022-06-24
Make the ONNX export code accept an op interface rather than the enum itself

Craigacp:
### Description
Introduces the `ONNXOperator` interface to allow users to insert custom ONNX ops, or ones using different opsets than opset 13 which is the base version Tribuo supports. The existing `ONNXOperators` enum now implements `ONNXOperator`, and in a future version of Tribuo we'll rename this to `ONNXOpset13` and split out the ONNX ML operators to a separate enum.

### Motivation
Currently Tribuo is limited to the ops available in `ONNXOperators`, this makes the ONNX model construction code insufficiently flexible.

### DIFF
--- a/Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
+++ b/Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
@@ -21,6 +21,7 @@ import ai.onnx.proto.OnnxMl;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Logger;
 
 import static org.tribuo.util.onnx.ONNXAttribute.VARIADIC_INPUT;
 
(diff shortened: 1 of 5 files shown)


Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
```diff
@@ -0,0 +1,228 @@
...
+        int numInputs = getNumInputs();
+        int numOptionalInputs = getNumOptionalInputs();
+        int numOutputs = getNumOutputs();
+        String opName = getOpName();
+        String domain = getOpDomain();
+        Map<String, ONNXAttribute> attributes = getAttributes();
+        Set<String> mandatoryAttributeNames = getMandatoryAttributeNames();
+
+        if ((numInputs != VARIADIC_INPUT) && ((inputs.length < numInputs) || (inputs.length > numInputs + numOptionalInputs))) {
+            throw new IllegalArgumentException("Expected " + numInputs + " inputs, with " + numOptionalInputs + " optional inputs, but received " + inputs.length);
```
JackSullivan, line 194: These error messages would be much more helpful in debugging if they contained the `opName`, `domain` and maybe also the expected number of input and output arguments. Perhaps they should all start with something like:
`String.format("For %s:%s(%d(+%d)) -> %d", domain, opName, numInputs numOptionalInputs, numOutputs)`

Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
```diff
@@ -0,0 +1,228 @@
...
+        } else if ((numInputs == VARIADIC_INPUT) && (inputs.length == 0)) {
+            throw new IllegalArgumentException("Expected at least one input for variadic input, received zero");
+        }
+        if (outputs.length != numOutputs) {
+            throw new IllegalArgumentException("Expected " + numOutputs + " outputs, but received " + outputs.length);
+        }
+        if (attributeValues.size() > attributes.size()) {
+            throw new IllegalArgumentException("Found more attributes than expected, received " + attributeValues.size() + ", expected at most " + attributes.size());
```
JackSullivan, line 202: Of all of the attribute correctness checks, this is the one that gives the least informative error message. Since these are short-circuiting, maybe this one should go after the other two.

Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
JackSullivan: Alternatively, if, from the domain and opName, we can reliably generate a link to the actual ONNX docs for that function, we should probably try to do so.

JackSullivan (commented):
Just some minor clarification of error messages in `ONNXOperator.build` and it should be good to go.

Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
Craigacp: We could generate a link if we recorded the version of that operator, which we don't really have (as the opset only updates some of the ops in each version). So I'd consider that future work. I'll fix the string to include the op name as you suggest.

Util/ONNXExport/src/main/java/org/tribuo/util/onnx/ONNXOperator.java
Craigacp: I removed this check as it's redundant given the one immediately after it.

JackSullivan (approved):
Looks good to me

PR #247, opened 2022-07-06
Added a Regex preprocessor

katieyounglove:
### Description
Added a simple implementation of the DocumentPreprocessor. The RegexPreprocessor applies regular expressions to a document in order and replaces any matches within the document with the corresponding replacement string.


Craigacp (approved):
LGTM

PR #254, opened 2022-08-02
Add classification feature selection support

Craigacp:
### Description
Adds support for feature selection algorithms via a `FeatureSelector` interface, a `SelectedFeatureSet` class to represent the output of a `FeatureSelector` and a `SelectedFeatureDataset` which applies a feature set to another dataset. It also includes implementations of 4 information theoretic feature selection algorithms (MIM, CMIM, JMI, mRMR) for classification tasks. The `FeatureSelector` interface is sufficiently general that we can implement it for regression, anomaly detection etc, though it does not cover operating on sequence data except via the `SequenceDataset.getFlatDataset` hook which flattens the sequence data into a series of independent examples. In the future we'll look at adding feature selection algorithms for regression and other output types, and before the next release we'll add a tutorial on the feature selection algorithms added in this PR.

The information theoretic algorithms bin the feature values using equal width bins to convert the real valued features into categoricals. We could add extensions to use the categorical values if the feature map believes them to be categorical, or to add other binning systems. This also densifies the dataset while it transposes it into a columnar view from Tribuo's default row view. We could use a sparse implementation of the columnar view which treats all features as binary, this has worked well in the past but the code is rather complicated so this initial version doesn't contain it.

Currently the feature selection algorithms are lightly tested to compare the parallel versions against the serial version. I've tested them offline against the implementations from [FEAST](https://github.com/Craigacp/FEAST) which is the reference implementation I developed in grad school. I'm considering adding a generated dataset with known outputs from FEAST to compare against in the unit tests.

### Motivation
Feature selection algorithms are useful for reducing the complexity of downstream prediction tasks, and for analysing the relevant features in a dataset.

### Paper reference
An overview of the feature selection techniques implemented here can be found in:

Brown G, Pocock A, Zhao M-J, Lujan M.
Conditional Likelihood Maximization: a Unifying Framework for Information Theoretic Feature Selection, JMLR 2012 - [PDF](https://www.jmlr.org/papers/volume13/brown12a/brown12a.pdf)


Core/src/main/java/org/tribuo/evaluation/EvaluationAggregator.java
```diff
@@ -174,6 +174,27 @@ public static <T extends Output<T>, R extends Evaluation<T>> Map<MetricID<T>, De
...
     }
 
+    /**
+     * Summarize all fields of a list of evaluations produced by {@link CrossValidation}.
+     * @param evaluations The evaluations to summarize.
+     * @param <T> The output type.
+     * @param <R> The evaluation type.
+     * @return The descriptive statistics for each metric.
+     */
+    public static <T extends Output<T>, R extends Evaluation<T>> Map<MetricID<T>, DescriptiveStats> summarizeCrossValidation(List<Pair<R, Model<T>>> evaluations) {
```
Craigacp, line 184: This change is separate from the rest of them but I added it to make it easier to aggregate across cross validation folds which is currently a little ugly to do and it didn't feel worth a full PR.

Core/src/main/java/org/tribuo/SelectedFeatureSet.java
```diff
@@ -0,0 +1,165 @@
...
+        return isOrdered;
+    }
+
+    @Override
+    public FeatureSetProto serialize() {
+        return ProtoUtil.serialize(this);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
JackSullivan, line 145: This notion of equality isn't immediately intuitive. We should document it.

Core/src/main/java/org/tribuo/dataset/SelectedFeatureDataset.java
```diff
@@ -0,0 +1,358 @@
...
+        DatasetDataCarrier<?> carrier = DatasetDataCarrier.deserialize(proto.getMetadata());
+        Class<?> outputClass = carrier.outputFactory().getUnknownOutput().getClass();
+        FeatureMap fmap = carrier.featureDomain();
+        List<Example<?>> examples = new ArrayList<>();
+        for (ExampleProto e : proto.getExamplesList()) {
+            Example<?> example = Example.deserialize(e);
+            if (example.getOutput().getClass().equals(outputClass)) {
+                for (Feature f : example) {
+                    if (fmap.get(f.getName()) == null) {
+                        throw new IllegalStateException("Invalid protobuf, feature domain does not contain feature " + f.getName() + " present in an example");
```
JackSullivan, line 204: Adding the index of the example would make debugging this much easier when thrown

Classification/FeatureSelection/src/main/java/org/tribuo/classification/fs/FSMatrix.java
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Calculates the joint mutual information I(X_i,X_j;Y) where Y is the label
+     * @param featureIndex The index i
+     * @param jointIndex The index j
+     * @return I(X_iX_j;Y)
+     */
+    double jmi(int featureIndex, int jointIndex);
+
+    /**
+     * Calculates the joint mutual information I(X_iX_j;X_k)
```
JackSullivan, line 109:
```suggestion
     * Calculates the joint mutual information I(X_i,X_j;X_k)
```

Classification/FeatureSelection/src/main/java/org/tribuo/classification/fs/FSMatrix.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param firstIndex The index i
+     * @param jointIndex The index j
+     * @param targetIndex The index k
+     * @return I(X_iX_j;X_k)
+     */
+    double jmi(int firstIndex, int jointIndex, int targetIndex);
```
JackSullivan, line 115:
```suggestion
     * @param secondIndex The index j
     * @param targetIndex The index k
     * @return I(X_i,X_j;X_k)
     */
    double jmi(int firstIndex, int secondIndex, int targetIndex);
```

Classification/FeatureSelection/src/main/java/org/tribuo/classification/fs/FSMatrix.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Generates an array of I(X_iZ;Y), ranging X_i over all possible features, where Y is the label.
+     * @param jointIndex The index of the Z variable
+     * @return An array where array[i] = I(X_iZ;Y)
```
JackSullivan, line 120:
```suggestion
     * Generates an array of I(X_i,Z;Y), ranging X_i over all possible features, where Y is the label.
     * @param jointIndex The index of the Z variable
     * @return An array where array[i] = I(X_i,Z;Y)
```

Classification/FeatureSelection/src/main/java/org/tribuo/classification/fs/FSMatrix.java
```diff
@@ -0,0 +1,191 @@
...
+     * Generates an array of I(X_iZ;Y), ranging X_i over all possible features.
+     * @param targetIndex The index of the Y variable
+     * @param jointIndex The index of the Z variable
+     * @return An array where array[i] = I(X_iZ;Y)
+     */
+    default double[] jmiList(int targetIndex, int jointIndex) {
+        int numFeatures = getNumFeatures();
+        double[] row = new double[numFeatures];
+        for (int i = 0; i < numFeatures; i++) {
+            row[i] = jmi(i,jointIndex,targetIndex);
```
JackSullivan, line 141: Why does this switch the argument order and naming from the previous version?
```suggestion
     * Generates an array of I(X_i,X_j;Y), ranging X_i over all possible features.
     * @param secondIndex The index of the X_j variable
     * @param targetIndex The index of the X_k variable
     * @return An array where array[i] = I(X_i,X_j;X_k)
     */
    default double[] jmiList(int secondIndex, int targetIndex) {
        int numFeatures = getNumFeatures();
        double[] row = new double[numFeatures];
        for (int i = 0; i < numFeatures; i++) {
            row[i] = jmi(i,secondIndex,targetIndex);
```

Classification/FeatureSelection/src/main/java/org/tribuo/classification/fs/FSMatrix.java
```diff
@@ -0,0 +1,191 @@
...
+     * Generates an array of I(X_i;Y|Z), ranging X_i over all possible features.
+     * @param targetIndex The index of the Y variable
+     * @param conditionIndex The index of the Z variable
+     * @return An array where array[i] = I(X_i;Y|Z)
+     */
+    default double[] cmiList(int targetIndex, int conditionIndex) {
+        int numFeatures = getNumFeatures();
+        double[] row = new double[numFeatures];
+        for (int i = 0; i < numFeatures; i++) {
+            row[i] = cmi(i,targetIndex,conditionIndex);
```
JackSullivan, line 187: Inconsistent naming between this and the singleton version
```suggestion
     * Generates an array of I(X_i;X_j|X_k), ranging X_i over all possible features.
     * @param featureIndex The index of the X_j variable
     * @param conditionIndex The index of the X_k variable
     * @return An array where array[i] = I(X_i;X_j|X_k)
     */
    default double[] cmiList(int featureIndex, int conditionIndex) {
        int numFeatures = getNumFeatures();
        double[] row = new double[numFeatures];
        for (int i = 0; i < numFeatures; i++) {
            row[i] = cmi(i,featureIndex,conditionIndex);
```

tutorials/feature-selection-tribuo-v4.ipynb
```diff
@@ -0,0 +1,539 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "bb95d7c0",
+   "metadata": {},
+   "source": [
+    "# Feature Selection\n",
+    "\n",
+    "When working with tabular machine learning datasets it is quite common to collect as many columns as possible, and extract as many features from those columns. This can lead to very large feature spaces, which can cause issues when fitting certain kinds of ML models, as they may use irrelevant or noisy features which lead to poor test time performance. Additionally large feature spaces can increase training times, while making it tricky to build interpretable models as there are too many features to reason over. For these reasons the field of feature selection developed, which aims to find the relevant subset of features for a given prediction task from a large set of possible features.\n",
```
JackSullivan, line 10:
```suggestion
    "When working with tabular machine learning datasets it is quite common to collect as many columns as possible, and extract many features from those columns. This can lead to very large feature spaces, which can cause issues when fitting certain kinds of ML models, as they may use irrelevant or noisy features leading to poor test time performance. Additionally, large feature spaces can increase training times, and make it difficult to build interpretable models as there are too many features to reason over. For these reasons the field of feature selection developed, which aims to find the relevant subset of features for a given prediction task from a large set of possible features.\n",
```

JackSullivan (changes requested):
Looks good apart from a few minor documentation quibbles

JackSullivan (approved):
Looks good to me

Mohammed-Ryiad-Eiadeh:
It would be awesome if you make a tutorial about that, and I am planning to
creat a wrapper model for feature selection using evolutionary algorithm.
If it's okay tell me please?

Craigacp:
The `summarizeCrossValidation` method will convert a list of evaluations from the `CrossValidation` class into a `Map<MetricID<T>, DescriptiveStats>` which gives you the mean, min, max & std deviation of each metric across the folds. You could then use that to drive an evolutionary procedure if you wanted. Alternatively, if you're asking how to implement the `FeatureSelector` interface for an evolutionary algorithm, then you can see how to use it in the feature selection tutorial added in this PR, but we don't have detailed docs on implementing the interface. Probably best to look at the existing implementations and open an issue if you get stuck.

Mohammed-Ryiad-Eiadeh:
Sounds well.
Cross Validation is suitable for the evaluation function when working with wrapper FS, and implementing the FS interface is a very good idea, but how about using TRIBUO to read data, evaluate the generated subsets and using TABLESAW to drop the noisy columns based on the used evolutionary algorithm and group all of that in a new module in TRIBUO library where this module can contain multiple evolutionary algorithms like Crow search, Black Hole, ...etc. the implementation will be based on stream, parallel stream, Runnable, Callable, Executor Service.
and I am sorry to bother you, Adam.

Craigacp:
The feature selection machinery will drop the unselected features for you, that's what `SelectedFeatureDataset` does. So you should only need to implement the evolutionary algorithms as `FeatureSelector` implementations which accept a `Trainer` and a number of folds (and any other parameters for the evolutionary part).

Mohammed-Ryiad-Eiadeh:
sounds very good, thank you

PR #255, opened 2022-08-08
Protobuf serialization for Output, OutputFactory and OutputInfo

Craigacp:
### Description
This is the next chunk of protobuf serialization support, adding support for Output, OutputInfo, OutputFactory and their subclasses. Most of this PR is the 15k+ lines of generated protos, the handwritten code & proto schemas is a lot smaller.

### Motivation
We need to move off Java serialization. See #226.


AnomalyDetection/Core/src/main/java/org/tribuo/anomaly/AnomalyFactory.java
```diff
@@ -51,6 +54,32 @@ public final class AnomalyFactory implements OutputFactory<Event> {
...
+            throw new IllegalArgumentException("Unknown version " + version + ", this class supports at most version " + 0);
+        }
+        if (message.getValue() != ByteString.EMPTY) {
+            throw new IllegalArgumentException("Invalid proto");
+        }
+        return new AnomalyFactory();
+    }
+
+    @Override
+    public OutputFactoryProto serialize() {
```
pogren, line 79: you annotated this class with:

`@ProtoSerializableClass(version = 0)`

Then you could use the default serialization:

```
    @Override
    public OutputFactoryProto serialize() {
        return ProtoUtil.serialize(this);
    }

```

pogren (approved):
Output, OutputFactory, OutputInfo now extend ProtoSerializable<OutputProto>
Major changes include:
- everything has serialization and deserialization now - either using default/built-in mechanism when possible, or custom serialization when not - nothing tricksy in here that I could see.
- A number of constructors were added to support deserialization - some of them with improved constraint checking
- various equals and hashCode methods added presumably to support simplified unit testing of equals with original objects with their serialized then deserialized twins

My only quibble is that LabelFactory, AnomalyFactory, ClusteringFactory, MultiLabelFactory could use the default serialization with an annotation like this: @ProtoSerializableClass(version = 0).  I can submit a PR if you want my local code for this.

Craigacp:
Sure. This PR isn't consistent with the others wrt the use of a static final `CURRENT_VERSION` field either but we can catch that up later.

PR #256, opened 2022-08-08
Bumping to ORT 1.12.1

Craigacp:
### Description
Bumps ONNX Runtime to the latest version which includes Linux & macOS arm64 support. It also enables the various ONNX tests on those platforms.

### Motivation
ARM support is very useful.


pogren (dismissed):
Update to ORT from version 1.9.0 to 1.12.1.  Adam says: "There are 3 changes which are roughly orthogonal. The first is in 1.11 I had to make a breaking change to the OrtEnvironment class in ONNX Runtime, and so the close method doesn’t throw anymore (as it’s now a no-op), so in Tribuo we had to remove the catch. The second is due to the longer lifetime of the OrtEnvironment there are now more warnings when running the tests about changing the name of the environment, so I removed all the places where we set it to a non-default value. The third change is allowing the ONNX tests to run on arm platforms as MS now include macOS and Linux arm binaries in their Maven Central download."  The code changes look consistent with each respectively.  I built this branch on my local Windows laptop and a remote linux server without a problem.  I also ran unit test coverage analysis tool for the unit tests in tribuo-classification-liblinear, tribuo-classification-libsvm, and tribuo-classification-sgd and the code in OnnxTestUtils.onnxLabelComparison executed as expected.

pogren (approved):
coverage analysis shows that the code in OnnxTestUtils is executed as expected when these tests are run on Windows - so it looks good to me.

PR #259, opened 2022-08-08
Bumping OpenCSV to 5.6

Craigacp:
### Description
Bumps to the latest version of OpenCSV.

### Motivation
Newer versions are better.


pogren (approved):
updates opencsv from version 5.4 to 5.6 and excludes transitive dependency on apache commons-lang3.  Builds on local Windows laptop and remote linux server without a problem.

PR #262, opened 2022-08-21
Protobuf serialization for math types

Craigacp:
### Description
Adds `ProtoSerializable` to all serializable types in `tribuo-math` and their subclasses. A more detailed description of the overall protobuf plan is given in #226.

### Motivation
We need to move off Java Serialization.


pogren (approved):
- Parameters, Kernel, VectorNormalizaer, Tensor, etc. interfaces are now ProtoSerializable and a bunch of impls implement serialize and deserializeFromProto
- added a bunch of equals, hashCode, constructors, and factory methods to support this
- added Util.product method

Everything here looks straightforward and much of the code has reasonable unit tests.  A few classes e.g. CRFParameter and FMParameters could use some unit testing for the proto serializable.

Craigacp:
I plan to add serialization unit tests for CRFModel when we get to that point which will necessarily test CRFParameters rather than individually unit testing CRFParameters. Ditto for FMParameters.

PR #264, opened 2022-08-31
Protobuf serialization for dataset, ensemble combiner, example and prediction

Craigacp:
### Description
Adds protobuf serialization for dataset, ensemble combiner, example, prediction, the sequence versions, and their implementations. Uses a carrier object for Dataset to make it easier to work with the metadata fields present in all datasets, and introduces a similar carrier for use in Model.

### Motivation


pogren (approved):
straightforward protobuf serialization code.  not much to say.  there are a few places where it might be nice to avoid duplicate code (e.g. deserializeFromProto in the subclasses SequenceDataset all populate the examples with the same 15 lines of code which could be put in a protected static method.)  But otherwise, this is straightforward and tested.

PR #266, opened 2022-09-02
reduced duplicate code by creating deserializeExamples methods in

pogren:
### Description
Reduce redundant code by creating a `deserializeExamples` method in both `Dataset` and `SequenceDataset` classes as a protected static method that the respective impls can use instead of repeating code.  
### Motivation
This is a cosmetic change to simplify code maintenance by consolidating redundant code. 

This small change came out of the review for PR #264.

### DIFF
--- a/Core/src/main/java/org/tribuo/MutableDataset.java
+++ b/Core/src/main/java/org/tribuo/MutableDataset.java
@@ -26,6 +26,7 @@ import org.tribuo.protos.core.ExampleProto;
 import org.tribuo.protos.core.MutableDatasetProto;
 import org.tribuo.provenance.DataProvenance;
 import org.tribuo.provenance.DatasetProvenance;
+import org.tribuo.sequence.SequenceExample;
 import org.tribuo.transform.TransformerMap;
 
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


Core/src/main/java/org/tribuo/MutableDataset.java
```diff
@@ -26,6 +26,7 @@
 import org.tribuo.protos.core.MutableDatasetProto;
 import org.tribuo.provenance.DataProvenance;
 import org.tribuo.provenance.DatasetProvenance;
+import org.tribuo.sequence.SequenceExample;
```
Craigacp, line 29: SequenceExample import is unnecessary.

Core/src/main/java/org/tribuo/sequence/SequenceDataset.java
```diff
@@ -211,5 +213,24 @@ public FlatDataset(SequenceDataset<T> sequenceDataset) {
         }
     }
 
+    protected static List<SequenceExample<?>> deserializeExamples(java.util.List<org.tribuo.protos.core.SequenceExampleProto> examplesList, Class<?> outputClass, FeatureMap fmap) {
```
Craigacp, line 216: Can you remove the fully qualified types & rename this to `deserializedSequenceExamples`?

Craigacp (changes requested):
Looks good. Just a few imports to tidy up.

Craigacp (approved):

PR #267, opened 2022-09-03
Protobuf serialization for MultinomialNaiveBayesModel

Craigacp:
### Description
The first of *many* protobuf serialization PRs for models. This is the simplest model so was straightforward.

### Motivation
We want to move off Java serialization.


Craigacp:
Most of this PR is a regeneration of the Math protos as they had the wrong proto namespace which caused weird runtime issues when depending on them from elsewhere.

JackSullivan (approved):
Looks good to me

PR #269, opened 2022-09-09
Adding SLM protobuf serialization

Craigacp:
### Description
Adds protobuf serialization support to sparse linear models.

### Motivation


pogren (approved):
straightforward proto serialization code is tested with reasonable code coverage.

PR #270, opened 2022-09-09
Adds XGBoost protobuf serialization.

Craigacp:
### Description
Adds protobuf serialization to XGBoost models.

### Motivation


pogren (approved):
straightforward proto serialization builds on linux and windows and test coverage is reasonable for the code added

PR #271, opened 2022-09-09
Adds protobuf serialization for OCI, ONNX and TF models

Craigacp:
### Description
Adds protobuf serialization to external models, and also some equals methods with thresholds for comparing `Output` and `Prediction`.

### Motivation


pogren (approved):
straightforward protobuf serialization code plus a few 'fullEquals' and 'distributionEquals' methods which take a threshold parameter

PR #272, opened 2022-09-09
Add protobuf serialization for LibSVM based models

Craigacp:
### Description
Add protobuf serialization for LibSVM.

### Motivation


pogren (approved):
straightforward pb serialization with good test code coverage.

PR #273, opened 2022-09-10
Adds protobuf serialization for LibLinear models

Craigacp:
### Description
Adds protobuf serialization for liblinear models. Unfortunately rebuilding a LibLinear Model requires reflective access into the model as the fields are all package private with no setters, so we resort to using a `ByteArrayOutputStream` to serialize the model using Java serialization before packing it into a protobuf. The protobufs do exist for doing it with reflection, but it'll cause issues with the module system so it's not turned on.

### Motivation
We need to move Tribuo off Java serialization.


pogren (approved):
straightforward protobuf serialization has good test code coverage.  I don't understand the purpose of the new methods in LibLinearModel: serializeModel, deserializeModels, and setField which have zero references and zero test code coverage.

PR #275, opened 2022-09-12
Adds SGD protobuf serialization

Craigacp:
### Description
Adds protobuf serialization to the SGD models (CRF, FM, linear, Kernel SVM).

### Motivation


pogren (approved):
looks good to me.  code test code coverage.

PR #276, opened 2022-09-12
Adds clustering & nearest neighbour protobuf serialization

Craigacp:
### Description
Adds protobuf serialization to HDBSCAN, K-Means, K-NN, and weighted ensemble models.

### Motivation


pogren (approved):
looks good to me overall.  My only nit to pick here is that it doesn't seem like WeightedEnsembleModel.serialize and deserializeFromProto are tested.

Craigacp:
Yeah, at the moment they aren't because I want to test them using a variety of models as the ensemble members and when this PR was made only the multinomial naive Bayes model was ProtoSerializable. It's on the to-do list once all the other PRs have landed.

PR #278, opened 2022-09-15
Adds protobuf serialization to Tree models

Craigacp:
### Description
Adds protobuf serialization to Tribuo's tree models. Unlike the other protobuf models this requires some complex logic to linearise the tree rather than arbitrarily nesting the protobufs. This is because in the future we will want to store the tree linearly anyway as it's *much* more cache efficient, and so having the storage format line up with that will make the internal transformation much simpler.

### Motivation


pogren (approved):
Looked this over pretty carefully and, in fact, tried to reimplement TreeModel.deserializeFromProtos because it seemed unnecessarily tortured.  Ended up understanding why it was that way and added a comment explaining the method.  Good test coverage.

pogren (approved):

PR #280, opened 2022-09-16
External proto consolidated equals

pogren:
### Description
consolidates fullEquals and distributionEquals so that the methods without a threshold param delegate to the ones that do instead of repeating code.

### Motivation
small attempt to reduce redundant code.


Craigacp (approved):
LGTM

PR #282, opened 2022-09-22
Adds documentation about the new protobuf serialization format and updates the helper programs to use it (updated)

Craigacp:
### Description
Adds some serde helper methods to `Model`, `Dataset`, `SequenceModel` and `SequenceDataset`. Adds documentation about the protobuf serialization to the markdown docs. Adds protobuf support to the various helper programs. Also adds a bunch more documentation for all the things added since 4.2.

### Motivation


pogren (approved):
javadoc comments, cleaned up imports, drivers have protobuf args, getTypeWitness in ObjectFactory, convenience methods for serialization/deserialization, cleaned up thrown exceptions - all looks straightforward and good to me.

PR #285, opened 2022-09-30
Migrates `DistanceType` to a `Distance` interface

### LINKED ISSUE #284: Limitid extendability by using Enums
Hi there,

I was really happy to find a Java-HDBSCAN-Implementation, but at the same time, I'm a little sad, why there are Enums used for the initialization? DistanceType and NeighboursQueryFactoryType are both enums and therefore not extendable with custom Metrics.
https://github.com/oracle/tribuo/blob/04c6fc35f95ec3843c0e2df3da88b91554c1abc1/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java#L197
For my usecase, I have a custom distance-Matric and also a matching custom Datastructure to provide nearest neighbours (it uses the distance between shapes), but I'd need to change the enum and can't just implement a simple interface and inject those custom parts.
Therefore, I think it would be good to migrate those enums to implement an interface that is used for the algorithm; then I could implement the interface and also use my code in the algorithm.

Craigacp:
### Description
Adds a `Distance` interface in `org.tribuo.math.distance`, along with implementations for Cosine, L1 and L2. Users can supply their own distance computation function by implementing the interface. All models which were previously migrated to use `DistanceType` now store a `Distance` instance, and the enum holds a singleton of each of the built in types for easy use in CLI programs and as default arguments.

cc @geoffreydstewart 

### Motivation
The set of available distances was too limiting, fixes #284.


geoffreydstewart:
@Craigacp This is a great improvement without breaking compatibility. Nice.

jhalexand (approved):
LGTM!

brainbytes42:
Well done, thanks!

PR #287, opened 2022-10-05
Fixes for test failures when running in a path with spaces in

Craigacp:
### Description
Due to iffy URL handling to get around some Windows issues when accessing test resources a bunch of failures happen if Tribuo is checked out into a path with spaces. These are all fixed by not using Strings as an intermediate step for paths.


pogren (approved):
I checked out main into a dir with a space in it and watched it fail.  Ran it again with the fixes and it built fine.  Changes look straightforward and reasonable.

PR #288, opened 2022-10-07
Bumping versions of XGBoost, jackson, protobuf and OCI

Craigacp:
### Description
We want to use the latest versions of these things.


jhalexand (approved):
LGTM

PR #309, opened 2022-12-19
FeatureHasher should have an option to not hash the values

### LINKED ISSUE #307: Negative values for term counts when using feature hashing
For a text classification task, I'm using feature hashing in a `TokenPipeline` to map the features (uni and bigrams) down to a smaller space. Then I want to apply a IDF transformation to them.
However this causes the following Exception:
````
java.lang.IllegalArgumentException: Example contained a NaN feature
````
How can I fix this?

I'm using the TokenPipeline like this: `var pipeline = new TokenPipeline(new UniversalTokenizer(), 2, true, maxNrFeatures)`.
When inspecting the values (term counts) for the features after extraction, I can see that some of them have negative values for some reason. I assume this causes the NaN values after applying the `IDFTransformation` to them.
Why are there negative values for term counts after using feature hashing?

**Is your question about a specific Tribuo class?**
`TokenPipeline`

**System details**
- Tribuo version 4.3.0

Craigacp:
### Description
`FeatureHasher` now has an option to preserve the values of the features that it hashes, and this option is available through `TokenPipeline`. Adds a couple of tests for the new behaviour.

This changes the default behaviour of `TokenPipeline` with hashing, and adds an option to the constructor to revert to the old behaviour. The old behaviour didn't make much sense, but this allows people to preserve their computations. Note that reproducing old models without this fix will turn on the new behaviour as it is defaulted to on, and so must be overridden when reproducing those models.

### Motivation
A term counting `TokenPipeline` should never return negative values for the term counts, but as `FeatureHasher` hashes the feature values into `{-1,1}` this could cause the hashed dimensions to aggregate to a negative value which causes problems elsewhere and semantically isn't term counting.

Fixes #307.


jhalexand (approved):
Looks good, with the minor nit that you document/describe (accurately) the value hash seed using the phrasing "value hash" but name the member variables and static final members using "hash value".  I think the former is clearer as "hash value" can easily be read as a more generic term.  I'm still approving since this is a very minor nit, but I'll not merge in case you want to change this first.

jhalexand (approved):
Looks good

PR #318, opened 2023-01-10
Add deserialization tests for 4.3 protobufs in AnomalyDetection, Clustering, MultiLabel

Craigacp:
### Description
Adds deserialization tests to AnomalyDetection/{Core, LibLinear, LibSVM}, Clustering/{Core, Hdbscan, Kmeans} and MultiLabel/{Core, SGD} for protobufs generated with v4.3.1. Also adds some helper methods for building the tests, and methods which generate the protobufs as we'll want to do this again when v5 is ready.

There are ~140 classes which are `ProtoSerializable` and we need coverage of all of them before moving on with the v5 refactoring. This PR covers 20 of the predictor specific types, and transitively a few things in `Core` which we'll add specific tests for too.

### Motivation
We will support 4.3 protobufs in v5 and need tests to ensure that the refactorings don't break serialization compatibility.


AnomalyDetection/LibLinear/src/test/java/org/tribuo/anomaly/liblinear/LibLinearAnomalyTrainerTest.java
```diff
@@ -69,4 +79,33 @@ public void gaussianDataTest() {
...
 
+    @Test
+    public void loadProtobufModel() throws IOException, URISyntaxException {
+        Path path = Paths.get(LibLinearAnomalyTrainerTest.class.getResource("liblinear-anomaly-431.tribuo").toURI());
+        try (InputStream fis = Files.newInputStream(path)) {
+            ModelProto proto = ModelProto.parseFrom(fis);
+            LibLinearAnomalyModel model = (LibLinearAnomalyModel) Model.deserialize(proto);
+
+            assertEquals("4.3.1", model.getProvenance().getTribuoVersion());
+
```
pogren, line 90: It seems we could be testing the inner model here with e.g.:

assertEquals(-1.0, model.getInnerModels().get(0).getBias());

I'm not sure if that is too in the weeds or not or maybe there's a good reason not to do this.

Craigacp: You mean we could explicitly check for model properties we know to be true in 4.3.1? I was considering comparing them against the freshly trained versions that exist elsewhere in the system, but for things like SGD there's some difference between ARM & x86_64 which would mean the tests need the tolerance parameters fudging to pass.

pogren (approved):
I ran the tests locally and looked them over.  It's not clear how one should regenerate the serialized files - but it's also not clear why one would want to. It's easy enough to add a "@Test" annotation and run them that way.  I did this with a couple of them and did a little inspection in the debugger.  It all seems reasonable to me.

PR #322, opened 2023-02-17
Add deserialization tests for 4.3 protobufs in Regression

Craigacp:
### Description
Following #318 adds deserialization tests to Regression and adds a missing class from MultiLabel.

### Motivation


pogren (approved):
This issue generates a bunch of new protobuf serialized models (in v4.3.1) and loads them back in in the unit tests and makes sure everything works - a kind of basic round trip for each model type's protobuf serialization.  It's pretty straightforward and the code builds and tests pass on my laptop.

PR #323, opened 2023-02-19
Fixes for protobuf creation in a few classes

Craigacp:
### Description
`DirectoryFileSource`, `SimpleStringDataSource` and `SimpleTextDataSource` all failed to call `super.getInstanceValues()` and so missed some values when converted into marshalled provenances. This PR relaxes the check for existing protobufs (as the missing value is a constant) and fixes `getInstanceValues` for these classes so it is correctly marshalled in the future. It also properly serializes `ClusterExemplar`s from v4.2 which may contain null values.

### Motivation
Protobuf serialization will be the only supported form in v5.


pogren (approved):
straightforward changes made to several classes to fix provenance value "short host name" plus a unit test.

PR #325, opened 2023-02-24
Upgrade TF-Java to 0.5.0

Craigacp:
### Description
Upgrades TF-Java to 0.5.0 and TF 2.10.

### Motivation
Keeping up to date with dependencies.


pogren (commented):
update to tensorflow requires very minimal code updates that look straightforward - no logic changes - just an updated return value from an api call.

PR #336, opened 2023-04-24
Refactor changes to improve code readability

prasanth08:
### Description
Removed  null checks before checking instanceof as null case is handled.
Reorder Access modifiers as per JSR standard
Use of lambda expression to improve readability.

### Motivation
Refactor changes to improve code readability.


Craigacp:
The null checks & access modifier reorderings are fine, but please don't modify the generated protobuf code (anything in a `protos` package) as those changes will be overwritten when we regenerate the protobufs as we change protobuf-java versions. In general we leave the public modifier in interfaces to make it explicit even for `default` methods, so don't remove that either.

prasanth08:
Got it! Made those chnages.

Craigacp:
Thanks for the contribution.

PR #345, opened 2023-07-13
Add deserialization tests for 4.3 protobufs in Classification

Craigacp:
### Description
Following https://github.com/oracle/tribuo/pull/318 adds deserialization tests to Classification.

### Motivation


pogren (approved):
looks like straightforward protobuf serialization testing to me.  Create the model, train it with some test data, write it out to a test resource.  Test is by deserializing the model and by recreating it and make sure both the deserialized and recreated models look the same and give the same results.  Repeat.  I didn't scrutinize every line of code - but it looks quite regular and straightforward to me.

PR #346, opened 2023-07-20
Add deserialization tests for 4.3 protobufs in Math

Craigacp:
### Description
Following https://github.com/oracle/tribuo/pull/318 adds deserialization tests to Math.

### Motivation


pogren (approved):
Looks good!  More straightforward code that generates provenance serializations and then tests them.  I made one small change that updated a few of the unit tests to use ParameterizedTest annotations with a MethodSource.

PR #356, opened 2023-11-03
Improve the determination of cluster exemplars

### LINKED ISSUE #355: HDBSCAN implementation in 4.3+
Hey,

Thanks for your work on this library and for making this open-source.

I was testing different implementations of HDBSCAN, and made a trivial test like this for Tribuo:

```
public void test() {
       final HdbscanTrainer trainer = new HdbscanTrainer(
                100,
                HdbscanTrainer.Distance.EUCLIDEAN,
                10,
                1
        );

        var outputFactory = new ClusteringFactory();

        var dataProvenance = new SimpleDataSourceProvenance(
                "test",
                outputFactory
        );

        var dataList = new ArrayList<Example<ClusterID>>();

        for (var n = 0; n < 10; n++) {
            for (double i = 1.0; i <= 2.0; i += 0.1) {
                dataList.add(createExample(i));
            }
            for (double i = 10.0; i <= 20.0; i += 0.1) {
                dataList.add(createExample(i));
            }
        }

        var dataSource = new ListDataSource(
                dataList,
                outputFactory,
                dataProvenance
        );

        var trainData = new MutableDataset<ClusterID>(
                dataSource
        );
        
        HdbscanModel model = trainer.train(trainData);
        System.out.printf("Num clusters: %d%n", new HashSet<>(model.getClusterLabels()).size());
        
        // Predicting on the same training data
        var predictions = model.predict(trainData);

        for (var prediction : predictions) {
            System.out.printf("%s: %d %f\n",prediction.getExample(), prediction.getOutput().getID(), prediction.getOutput().getScore());
        }
}
        
private Example<ClusterID> createExample(double value) {
        var example = new ArrayExample<ClusterID>(
                ClusteringFactory.UNASSIGNED_CLUSTER_ID,
                new String[]{"A"},
                new double[]{value}
        );
        return example;
}
```

Now surprisingly, version 4.2 worked as expected(?) identifying 2 clusters and prediction work as well.
For version 4.3 exactly the same code above identifies also 2 clusters, but the prediction returns all as outliers (ClusterID: 0  with score: 0.999900).

Am I missing something or HDBSCAN implementation in 4.3 gives unexpected results?

**Is your question about a specific ML algorithm or approach?**
HDBSCAN implementation in Tribuo.

**System details**
- Tribuo version: 4.2.2, 4.3.1

geoffreydstewart:
### Description
We need to improve the determination of cluster exemplars, which are used after the model is trained to make predictions. Some cases, such as those which use contrived datasets, can result in very "clean" clusters, with no outliers. One such case has exposed an issue in the current logic to determine cluster exemplars.

### Motivation
This change fixes #355


Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -778,31 +779,68 @@ private static List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Inte
...
+                    Stream<Integer> intStreamTree = IntStream.range(0, outlierScoreTreeSize).boxed();
+                    intStreamTree.forEach((i) -> {
+                        Entry<Double, Integer> treeEntry = outlierScoreIndexTree.pollFirstEntry();
+                        partialClusterExemplarsList.add(new Pair<>(treeEntry.getKey(), treeEntry.getValue()));
+                    });
+
+                    // To determine the remaining exemplars, the best thing to do is randomly sample them from all the
+                    // points in this cluster. This could introduce duplicate exemplar points, but that is safer than
+                    // reducing the number of exemplars.
+                    Random rand = new Random();
```
Craigacp, line 824: We can't use an unseeded random here, it's not reproducible and isn't captured in the provenance. I don't really want to break the API to fix a bug, so maybe here you could seed the random with a fixed seed (and use SplittableRandom rather than Random, it's much higher quality), then in `main` we make that seed a config parameter and default it to the seed you chose?

Clustering/Hdbscan/src/test/java/org/tribuo/clustering/hdbscan/TestHdbscan.java
```diff
@@ -19,12 +19,7 @@
 import com.oracle.labs.mlrg.olcut.util.Pair;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
-import org.tribuo.DataSource;
-import org.tribuo.Dataset;
-import org.tribuo.Feature;
-import org.tribuo.Model;
-import org.tribuo.MutableDataset;
-import org.tribuo.Prediction;
+import org.tribuo.*;
```
Craigacp, line 22: Could you flip this back to an explicit import here and for `java.util.*`?

geoffreydstewart: Yes, this was supposed to come in with the previous commit. The comment mentioned it but that doesn't really help.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
geoffreydstewart: Yes, thank you for pointing this out. For `main`, the idea is that we should prepare an additional, similar PR, but with a new attribute in the `HdbscanTrainer.java` perhaps: 
`private long exemplarSeed = 12345L` annotated with `Config`
and provide this attribute to the instantiation of `SplittableRandom`?

geoffreydstewart:
The author of #355 seemed motivated to test out the fix to this issue. Perhaps they have some other use cases in mind. I was wondering if it would be worthwhile to offer to share the source branch containing the fix, to see if there is any additional  feedback.

Craigacp (approved):
LGTM thanks.

PR #357, opened 2023-11-17
Improve the determination of cluster exemplars fix for main

### LINKED ISSUE #355: HDBSCAN implementation in 4.3+

geoffreydstewart:
### Description
We need to improve the determination of cluster exemplars, which are used after the model is trained to make predictions. Some cases, such as those which use contrived datasets, can result in very "clean" clusters, with no outliers. One such case has exposed an issue in the current logic to determine cluster exemplars. The seed, which is used when sampling cluster exemplars at random from members of a cluster, is exposed as a config parameter.

### Motivation
This change fixes #355 on the main branch.

### DIFF
--- a/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
+++ b/Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
@@ -155,7 +155,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
         "from the members of a cluster.")
     private long exemplarSampleSeed = Trainer.DEFAULT_SEED;
 
-    private SplittableRandom rng = new SplittableRandom(exemplarSampleSeed);
+    private SplittableRandom rng;
 
     private int trainInvocationCounter;
 
@@ -207,6 +207,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
         this.k = k;
         this.numThreads = numThreads;
         this.neighboursQueryFactory = NeighboursQueryFactoryType.getNeighboursQueryFactory(nqFactoryType, dist, numThreads);
+        postConfig();
     }
 
     /**
@@ -227,7 +228,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
         this.numThreads = numThreads;
         this.neighboursQueryFactory = NeighboursQueryFactoryType.getNeighboursQueryFactory(nqFactoryType, dist, numThreads);
         this.exemplarSampleSeed = exemplarSampleSeed;
-        this.rng = new SplittableRandom(exemplarSampleSeed);
+        postConfig();
     }
 
     /**
@@ -242,6 +243,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
         this.dist = neighboursQueryFactory.getDistance();
         this.k = k;
         this.neighboursQueryFactory = neighboursQueryFactory;
+        postConfig();
     }
 
     /**
@@ -268,6 +270,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
             }
         }
 
+        this.rng = new SplittableRandom(exemplarSampleSeed);
     }
 
     @Override
@@ -790,7 +793,7 @@ public final class HdbscanTrainer implements Trainer<ClusterID> {
      * @param rng The RNG to use.
      * @return A list of {@link ClusterExemplar}s which are used for predictions.
      */
-    private List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Integer, List<Pair<Double, Integer>>> clusterAssignments,
+    private static List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Integer, List<Pair<Double, Integer>>> clusterAssignments,
                                                           org.tribuo.math.distance.Distance dist, SplittableRandom rng) {
         List<ClusterExemplar> clusterExemplars = new ArrayList<>();
         // The formula to calculate the exemplar number. This calculates the number of exemplars to be used for this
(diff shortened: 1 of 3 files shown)


Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -779,31 +804,68 @@ private static List<ClusterExemplar> computeExemplars(SGDVector[] data, Map<Inte
...
+                    Stream<Integer> intStreamTree = IntStream.range(0, outlierScoreTreeSize).boxed();
+                    intStreamTree.forEach((i) -> {
+                        Entry<Double, Integer> treeEntry = outlierScoreIndexTree.pollFirstEntry();
+                        partialClusterExemplarsList.add(new Pair<>(treeEntry.getKey(), treeEntry.getValue()));
+                    });
+
+                    // To determine the remaining exemplars, the best thing to do is randomly sample them from all the
+                    // points in this cluster. This could introduce duplicate exemplar points, but that is safer than
+                    // reducing the number of exemplars.
+                    SplittableRandom rand = new SplittableRandom(exemplarSampleSeed);
```
Craigacp, line 849: How much of an edge case is this randomized fallback? We have special treatment for all other RNGs in Tribuo where they are members of the trainer and split under a lock to preserve provenance information, which ensures that repeated runs of a trainer on the same data give different answers when not controlling the RNG state (e.g. when used in an ensemble). If this is purely an edge case issue then it might be ok to leave it as is, but if it might occur relatively frequently then it's probably worth using the same idiom we do elsewhere.

geoffreydstewart: This edge case seems to be quite rare, but if there is a "best solution" which can be implemented here we should strive towards that. I'll assume that the implementation of this special treatment demonstrated in the `KMeansTrainer` is a good example to follow.

Craigacp: Yeah, this synchronized idiom along with storing an rng as a field - https://github.com/oracle/tribuo/blob/main/Clustering/KMeans/src/main/java/org/tribuo/clustering/kmeans/KMeansTrainer.java#L278.

geoffreydstewart: I've added a commit which adds a RNG as a field, and the supporting logic. Let me know if anything else might be needed.

Clustering/Hdbscan/src/main/java/org/tribuo/clustering/hdbscan/HdbscanTrainer.java
```diff
@@ -150,6 +151,12 @@ public DistanceType getDistanceType() {
     @Config(description = "The nearest neighbour implementation factory to use.")
     private NeighboursQueryFactory neighboursQueryFactory;
 
+    @Config(description = "The seed to use in the event cluster exemplars need to be determined with random samples " +
+        "from the members of a cluster.")
+    private long exemplarSampleSeed = Trainer.DEFAULT_SEED;
+
+    private SplittableRandom rng = new SplittableRandom(exemplarSampleSeed);
```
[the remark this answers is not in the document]
geoffreydstewart: Right, I've made this change as well as adding the call to `postConfig` from the Constructors.

geoffreydstewart:
> Needs a fix to the RNG creation, otherwise looks good.

Just following up to see what else might be needed here, no rush though.

Craigacp (approved):
LGTM, thanks.

PR #369, opened 2024-05-20
Gaussian Mixture Model implementation

### LINKED ISSUE #359: Gaussian Mixture Model capability
Hi, 

I need to port some Matlab code to java and I'm looking at what is out there in Java land that can do Gaussian Mixture Models.  Specifically, the code that I have to port is making heavy use of Matlab's gmdistribution  https://www.mathworks.com/help/stats/gmdistribution.html and fitgmdist https://www.mathworks.com/help/stats/fitgmdist.html.  I see that Tribuo alludes to Gaussian Mixtures in the KMeans tutorial: https://tribuo.org/learn/4.3/tutorials/clustering-tribuo-v4.html. So maybe this would suffice? I'm definitely not a mathematician, but I'm trying to see if Tribuo can do GMMs like these Matlab functions.  It appears that Matlab supports two covariance types 'full' and diagonal'.  

Can you please elaborate on Tribuo's capabilities in regards to GMMs?

Craigacp:
### Description
Adds a Gaussian Mixture Model clustering implementation with spherical, diagonal and full covariance structure using Expectation Maximisation. Also adds a mixture distribution to the RNG library to allow sampling from a user constructed gaussian mixture model (as opposed to one fit to a data distribution).

It also contains new Math functions necessary to implement the GMM efficiently, some updates for K-Means to modernise it a little bit, and some cleanups to the main pom file.

There are a few important fixes to the Math package in here as well, determinants for matrix factorizations were incorrectly computed, and the subtract function on SparseVector was incorrect.

### Motivation
GMMs are a useful clustering algorithm. Fixes #359.

### Paper reference
Hastie, T., Tibshirani, R., Friedman, J. H., & Friedman, J. H. (2009). The Elements of Statistical Learning: Data Mining, Inference, and Prediction, Second Edition.


Core/src/main/java/org/tribuo/util/Util.java
```diff
@@ -517,6 +537,25 @@ public static int sampleFromCDF(double[] cdf, Random rng) {
...
     }
 
+    /**
+     * Samples an index from the supplied cdf.
+     * @param cdf The cdf to sample from.
+     * @param rng The rng to use.
+     * @return A sample.
+     */
+    public static int sampleFromCDF(double[] cdf, RandomGenerator rng) {
+        if (Math.abs(cdf[cdf.length-1] - 1.0) > 1e-6) {
```
JackSullivan, line 547: Why to we have a different epsilon here vs. in validatePMF above?

Math/src/main/java/org/tribuo/math/distributions/MultivariateNormalDistribution.java
```diff
@@ -91,62 +99,278 @@ public MultivariateNormalDistribution(DenseVector means, DenseMatrix covariance,
...
             }
-        } else {
-            Optional<DenseMatrix.CholeskyFactorization> factorization = this.covariance.choleskyFactorization();
-            if (factorization.isPresent()) {
-                this.samplingCovariance = factorization.get().lMatrix();
-            } else {
-                throw new IllegalArgumentException("Covariance matrix is not positive definite.");
+            case SPHERICAL -> {
+                if (covariance instanceof DenseVector vec) {
+                    if ((vec.size() != 1) && (vec.size() != means.size())) {
```
JackSullivan, line 206: I'm confused by the second clause here - shouldn't the first one be the only one needed?

Math/src/main/java/org/tribuo/math/distributions/MultivariateNormalDistribution.java
```diff
@@ -91,62 +99,278 @@ public MultivariateNormalDistribution(DenseVector means, DenseMatrix covariance,
...
+    }
+
+    public static double logProbability(SGDVector input, DenseVector mean, Tensor precision, double determinant, CovarianceType type) {
+        // p(input|mean, variance) = \frac{1}{(2\pi)^{d/2} determinant^{1/2}} e^{-1/2 * (input - mean)^T * precision * (input - mean)}
+        // log p(i|mu,sigma) = - log ({2\pi}^{d/2}) - log (determinant^{1/2}) + (-1/2 * (i - mu)^T * precision * (i - mu))
+        double scalar = (- (mean.size() / 2.0) * Math.log(2 * Math.PI)) - (Math.log(determinant) / 2.0);
+        DenseVector diff = (DenseVector) input.subtract(mean);
+        double distance = switch (type) {
+            case FULL -> {
+                DenseMatrix precMat = (DenseMatrix) precision;
```
JackSullivan, line 298: I know we've done all the checking already in the construction of the object, but this static method is public and doesn't do any checking. Are we ok with throwing ClassCastExceptions here? It's ok if the answer is yes, but then this method should grow some documentation about being a bit sharp, maybe.

Clustering/GMM/src/main/java/org/tribuo/clustering/gmm/GMMTrainer.java
```diff
@@ -0,0 +1,617 @@
...
+                if (parallel) {
+                    try {
+                        normSum = fjp.submit(() -> zipEStream.parallel().mapToDouble(eStepFunc).sum()).get();
+                    } catch (InterruptedException | ExecutionException e) {
+                        throw new RuntimeException("Parallel execution failed", e);
+                    }
+                } else {
+                    normSum = zipEStream.mapToDouble(eStepFunc).sum();
+                }
+                logger.log(Level.FINE, i + "th e step completed.");
```
JackSullivan, line 301:
```suggestion
                logger.log(Level.FINE, i + "the E step completed.");
```

Math/src/main/java/org/tribuo/math/distributions/MultivariateNormalDistribution.java
Craigacp: It allows the covariance to be a vector of length means with all the same value. Though I'm not checking that they are the same value here.

Clustering/GMM/src/main/java/org/tribuo/clustering/gmm/GMMTrainer.java
```diff
@@ -0,0 +1,617 @@
...
+                    }
+                    case DIAGONAL, SPHERICAL -> {
+                        for (int j = 0; j < covariances.length; j++) {
+                            DenseVector covVec = (DenseVector) covariances[j];
+                            covVec.intersectAndAddInPlace(covarianceJitter);
+                            DenseVector preVec = (DenseVector) precision[j];
+                            double tmp = 1;
+                            for (int k = 0; k < preVec.size(); k++) {
+                                double curVal = 1/Math.sqrt(covVec.get(k));
+                                preVec.set(k, curVal);
```
JackSullivan, line 448: I know `precision[j]` gets set by mutation, but it might be nice to assign it anyway or have a comment for explicitness.

Core/src/main/java/org/tribuo/util/Util.java
Craigacp: I don't remember, I'll make them uniform. I think we use 1e-6 in most places for the zero check.

Math/src/main/java/org/tribuo/math/distributions/MultivariateNormalDistribution.java
Craigacp: Sure, I'll fix that.

Clustering/GMM/src/main/java/org/tribuo/clustering/gmm/GaussianMixtureModel.java
```diff
@@ -0,0 +1,394 @@
...
+        ClusterID max = null;
+        Map<String, ClusterID> scores = new HashMap<>();
+        for (int i = 0; i < distributions.length; i++) {
+            ClusterID tmp = new ClusterID(i, responsibilities.get(i));
+            if (max == null || tmp.getScore() > max.getScore()) {
+                max = tmp;
+            }
+        }
+
+        return new Prediction<>(max,scores,vector.size(),example,generatesProbabilities);
```
JackSullivan, line 297: does it make sense for `generatesProbabilities` to ever not be `true` here?

Craigacp: It's set to true in the constructor and the field is final, so it can't ever be false. I can change it to be a literal true if you want.

JackSullivan: I think I'd prefer that slightly, or at least a comment to that effect.

JackSullivan (changes requested):
Looks good apart from the above nitpicks

Core/src/main/java/org/tribuo/util/Util.java
Craigacp: So it's that in some cases the CDF can come from floats upcast into doubles, where the tolerance is less tight, but the validate function is used for doubles. I'll add some documentation.

Craigacp:
I've updated this PR after the review.

JackSullivan (approved):
Looks good to me

PR #372, opened 2024-06-12
fixed markdown of last table

behrica:
Did not render as table due to one separator column too much

### Description
Fix the markdown of a table

### Motivation
I saw that it rendered "badly", not as table


Craigacp:
Thanks for the PR, I should have spotted that when I added the last table so it's good to have it fixed. Once you've signed the OCA I can run our CI checks and merge this in.

Craigacp:
This was fixed in another PR.

PR #376, opened 2024-08-15
Fix #375

### LINKED ISSUE #375: SQLDataSource does not close connections
The SQLDataSource does not close connections and leaks them.
This goes well if it is invocated at a low frequency and the jdbc driver releases connections on finalize(). 
But, this is not always the case and will lead to the database either dropping connections or crashing. Either way, its not the proper way to handle JDBC resources.

The culprit is the getStatement() method in SQLDBConfig, which creates a statement but discards the connection.
Its only used in SQLDataSource.rowIterator(). The SQLDataSource should retrieve the connection and at least store it in the SQLDataSource.

SQLDataSource already implements AutoCloseable, so the connections can be closed at the end of close().

JackSullivan:
### Description

Added `SQLDBConfig::getStatementAndConnection` method and deprecated `SQLDBConfig::getStatement`, switching `SQLDatasource` to use the latter. This deals with a leak where `getStatement` wasn't closing DB connections

### Motivation

Fix #375


Craigacp (approved):
LGTM

Issue #383, opened 2024-11-26
IntArrayContainer merge method creates duplicates int values


gilleshuron:
**Describe the bug**

This method:

> org.tribuo.common.tree.impl.IntArrayContainer#merge(java.util.List<int[]>, org.tribuo.common.tree.impl.IntArrayContainer, org.tribuo.common.tree.impl.IntArrayContainer) 

... incorrectly merges list of int[], creates duplicated int values during the merge.
 
At the least, it impacts performance. Variance computing is completely wrong in the CARTRegressionTrainer.

**To Reproduce**

Create a unit test with this code:

```
public void bug() {
    IntArrayContainer first = new IntArrayContainer(10);
    IntArrayContainer second = new IntArrayContainer(10);
    List<int[]> list = List.of(
        new int[]{1, 2, 4},
        new int[]{3, 5, 6, 7, 8, 9, 10}
    );
    IntArrayContainer.merge(list, first, second); //<-- buggy method
    Assertions.assertEquals(1, first.array[0]);
    Assertions.assertEquals(2, first.array[1]); //fails !!! array is filed with [1,1,2,2,3,4,4,5,6,7,8,9,10,0,0]
    //expected is [1,2,3,4,5,6,7,8,9,10]

  }

```

**Expected behavior**

    //expected is [1,2,3,4,5,6,7,8,9,10]

**System information:**

 - Tribuo Version: [4.3.1]
 - OS: [Windows]
 - Java Version: [21]
 - JDK Vendor: [temurin-21.0.3]

**Additional context**

The mistake may lie in how the firstBuffer is initialized and updated in each subsequent merge:

```
public static int[] merge(List<int[]> input, IntArrayContainer firstBuffer, IntArrayContainer secondBuffer) {
    if (input.size() > 0) {
        firstBuffer.fill(input.get(0));
        for (int i = 1; i < input.size(); i++) { // Start from the second array
            merge(firstBuffer, input.get(i), secondBuffer);
            IntArrayContainer tmp = secondBuffer;
            secondBuffer = firstBuffer;
            firstBuffer = tmp;
        }
        return firstBuffer.copy();
    } else {
        return new int[0];
    }
}
```

Modification made:
    Changed the loop start from i = 0 to i = 1. The first array is already used to initialize firstBuffer, so the merging should start from the second array (i = 1).

Craigacp:
[...]

PR #384, opened 2024-11-27
Fixing a bug in IntArrayContainer.merge and adding tests

Craigacp:
### Description
The merge method should not duplicate inputs, which can lead to issues in the regression trees which use that method.

### Motivation
Fix for #383 and more tests for `IntArrayContainer`.

### DIFF
--- a/Common/Trees/src/main/java/org/tribuo/common/tree/impl/IntArrayContainer.java
+++ b/Common/Trees/src/main/java/org/tribuo/common/tree/impl/IntArrayContainer.java
@@ -88,7 +88,7 @@ public class IntArrayContainer {
      * @param other The array to copy from.
      */
     public void fill(IntArrayContainer other) {
-        if (other.array.length > array.length) {
+        if (other.size > array.length) {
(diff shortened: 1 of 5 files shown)


Common/Trees/src/main/java/org/tribuo/common/tree/impl/IntArrayContainer.java
```diff
@@ -88,7 +88,7 @@ public void fill(int[] otherArray) {
      * @param other The array to copy from.
```
JackSullivan, line 88: Might be good to explicitly note that this truncates the array as well as filling.

JackSullivan (dismissed):

Common/Trees/src/main/java/org/tribuo/common/tree/impl/IntArrayContainer.java
Craigacp: Fixed.

JackSullivan (approved):
Looks good to me

PR #389, opened 2025-01-10
Remove java.io.Serializable

Craigacp:
### Description
Removes support for java.io.Serializable from all main Tribuo types. The only supported serialization mechanism is protobuf.

### Motivation
java.io.Serializable makes it difficult to migrate classes.


Classification/Experiments/src/main/java/org/tribuo/classification/experiments/Test.java
```diff
@@ -134,33 +126,12 @@ public static Pair<Model<Label>,Dataset<Label>> load(ConfigurableTestOptions o)
...
-            } catch (ClassNotFoundException e) {
-                throw new IllegalArgumentException("Unknown class in serialised model", e);
-            }
-        }
+        Model<?> tmpModel = Model.deserializeFromFile(modelPath);
         Model<Label> model = tmpModel.castModel(Label.class);
         logger.info(String.format("Loading data from %s", datasetPath));
         Dataset<Label> test;
         switch (o.inputFormat) {
             case SERIALIZED:
```
JackSullivan, line 134: Isn't this kind of a breaking change? Code that previously expected java serialization would not inadvertently be using a proto serialization path.

Classification/SGD/src/main/java/org/tribuo/classification/sgd/crf/SeqTest.java
```diff
@@ -118,11 +114,6 @@ public String getOptionsDescription() {
          */
         @Option(longName = "model-hashing-salt", usage = "Salt for hashing the model.")
         public String modelHashingSalt = "";
-        /**
-         * Load in the model in protobuf format.
-         */
-        @Option(longName = "protobuf-model", usage = "Load the model from a protobuf. Optional")
```
JackSullivan, line 124: Again isn't this a breaking change? Code that was using this arg will now crash, even though it was doing things the 'right' way

Data/src/main/java/org/tribuo/data/DataOptions.java
```diff
@@ -217,30 +210,6 @@ public <T extends Output<T>> Pair<Dataset<T>,Dataset<T>> load(OutputFactory<T> o
         char separator;
         switch (inputFormat) {
             case SERIALIZED:
```
JackSullivan, line 212: Another breaking change, as above

Interop/OCI/src/main/java/org/tribuo/interop/oci/OCIModelCLI.java
```diff
@@ -266,11 +245,6 @@ public String getOptionsDescription() {
          */
         @Option(charName='s',longName="dataset-path",usage="Path to the serialized dataset to score.")
         public Path datasetPath;
-        /**
-         * Is the serialized dataset in protobuf format?
-         */
-        @Option(longName="dataset-protobuf",usage="Is the serialized dataset a protobuf?")
```
JackSullivan, line 272: ibid

Json/src/main/java/org/tribuo/json/StripProvenance.java
```diff
@@ -280,11 +275,6 @@ public String getOptionsDescription() {
          */
         @Option(charName = 't', longName = "hash-type", usage = "The hash type to use.")
         public ProvenanceUtil.HashType hashType = ObjectProvenance.DEFAULT_HASH_TYPE;
-        /**
-         * Read and write protobuf formatted models.
-         */
-        @Option(longName = "model-protobuf", usage = "Read and write protobuf formatted models.")
```
JackSullivan, line 286: Breaking change

JackSullivan (commented):
There are a couple of places here where behavior changes for people who were already using protobuf models. We should think about how to handle that.

Json/src/main/java/org/tribuo/json/StripProvenance.java
Craigacp: I've reverted this to allow this argument to be a no-op as StripProvenance is part of the supported API.

PR #401, opened 2025-04-23
Fix the Adam gradient optimizer

Craigacp:
### Description
The optimizer was multiplying by the squared gradient accumulator rather than dividing by it, this is fixed now.

### Motivation
Correcting the implementation of Adam.


JackSullivan (approved):
Looks good to me

PR #410, opened 2025-10-17
Add epsilon to impurity comparisons in tree construction

### LINKED ISSUE #407: "bounds must be positive" Exception during ExtraTreesTrainer training in certain nodes
**Describe the bug**
While training an ExtraTreesTrainer on a dataset with multiple features and high tree depth, the training process sometimes fails after successfully building hundreds of trees. The following exception is thrown:
```
bound must be positive
at org.tribuo.regression.rtree.impl.RegressorTrainingNode.buildRandomTree(RegressorTrainingNode.java:236)
at org.tribuo.regression.rtree.impl.RegressorTrainingNode.buildTree(RegressorTrainingNode.java:150)
```

**To Reproduce**
The following configuration was used:
```
var trainer = new ExtraTreesTrainer<>(
    new CARTRegressionTrainer(
        12,       // maxDepth: tree depth
        1.0f,     // minChildWeight
        0.0f,     // minImpurityDecrease
        0.5f,     // fractionFeaturesInSplit
        true,     // useRandomSplitPoints
        new MeanSquaredError(), // impurity
        42L       // seed
    ),
    new AveragingCombiner(),
    500,       // numTrees
    42L
);
```

**Additional context**

In RegressorTrainingNode.buildRandomTree:
```
// if there is only 1 inverted feature for this feature, it has only 1 value, so cannot be split
if (feature.size() == 1) {
    continue;
}

int splitIdx = rng.nextInt(feature.size() - 1);
```

**Observation / Hypothesis**
In certain nodes, there may be no values left for a selected feature.

In such cases, feature.size() == 0, which results in rng.nextInt(feature.size() - 1) → rng.nextInt(-1) → "bounds must be positive" exception.

The current check feature.size() == 1 only handles the case with exactly one value, not the empty case.

Craigacp:
### Description
Adds an epsilon to the impurity check when finding the best split in tree construction to prevent splits with an empty branch from comparing as lower than the node's impurity.

### Motivation
Fixes #407.

### DIFF
--- a/Classification/DecisionTree/src/main/java/org/tribuo/classification/dtree/impl/ClassifierTrainingNode.java
+++ b/Classification/DecisionTree/src/main/java/org/tribuo/classification/dtree/impl/ClassifierTrainingNode.java
@@ -162,7 +162,7 @@ public class ClassifierTrainingNode extends AbstractTrainingNode<Label> {
                 double lessThanScore = impurity.impurityWeighted(lessThanCounts);
                 double greaterThanScore = impurity.impurityWeighted(greaterThanCounts);
                 double score = (lessThanScore + greaterThanScore) / weightSum;
-                if (score < bestScore) {
+                if ((score + IMPURITY_EPSILON) < bestScore) {
                     bestID = i;
                     bestScore = score;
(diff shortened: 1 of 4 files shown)


jhalexand (approved):
Looks good!

PR #414, opened 2025-11-10
Upgrade XGBoost to 3.0.4

Craigacp:
### Description
Move XGBoost to 3.0.4 which is the latest version that supports the deprecated model format. We'll move to 3.1 or newer after there's a Tribuo release which allows users to migrate their models by saving them.

Models are now saved in the XGBoost `ubj` format inside Tribuo's protobuf based model serialization format, older models are saved in the old default string format which has been removed from XGBoost 3.1. To upgrade to the new format a model must be loaded in Tribuo and saved again.

It also bumps the versions of all the Maven plugins to latest, and pins the Maven version to 3.9 in prep for moving to Maven 4 when it is released.

Also adds a Linux ARM64 CI runner and updates to Java 25 from Java 24 in the CI.

### Motivation
Keep up with new dependencies, and to pull in macOS ARM64 support.

### DIFF
--- a/.github/workflows/maven-macos-aarch64.yml
+++ b/.github/workflows/maven-macos-aarch64.yml
@@ -19,6 +19,8 @@ jobs:
     name: macOS aarch64 Java SE ${{ matrix.java }}
     steps:
       - uses: actions/checkout@v4
+      - name: Install libomp for XGBoost
+        run: brew install libomp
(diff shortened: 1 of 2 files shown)


jhalexand (approved):
A whole smorgasbord!  LGTM

PR #419, opened 2025-11-21
Fix flaky test in TestCART by stabilizing HashMap iteration order

lbh930:
### Description
This PR fixes test flakiness in TestCART by making the test independent from HashMap ordering. This PR only affects test code and did not touch production code.

### Motivation
This change is necessary to reduce test flakiness caused by HashMap's non-deterministic iteration order. As observed by NonDex, the test would occasionally fail when HashMap key ordering varied across JVM runs. This is a bug fix that increases test stability.

### Paper reference
N/A


Craigacp:
#421 contains a fix for the root cause.

Craigacp:
Closed in favour of #421.

PR #420, opened 2025-11-21
Fix flaky test in TestClassificationEnsembles by stabilizing HashMap iteration order

lbh930:
### Description
This PR fixes non-deterministic test flakiness in TestClassificationEnsembles by stabilizing HashMap iteration order during feature map assertions. The fix adds a sortFeatures() helper method to sort feature maps before comparison.

### Motivation
This change is necessary to eliminate test flakiness caused by HashMap's non-deterministic iteration order in feature maps returned by getTopFeatures(). This is a bug fix that ensures test stability without modifying any production code. The fix is verified using NonDex.

### Paper reference
N/A


lbh930:
@Craigacp I think so, I just verified with NonDex that #424 fixes this test too. Thank you!

PR #422, opened 2025-11-29
Fix test-flakiness in MIMTest caused by floating point comparison

lbh930:
### Description
This PR fixes non-deterministic test flakiness in MIMTest by relaxing floating-point assertions with a small tolerance. The change compares each score using an epsilon (1e-12) to account for floating point variation caused by summation order.

### Motivation
NonDex found failures due to tiny floating-point discrepancies caused by iteration order perturbations. The problem lies in the test code only that: non-deterministic HashMap iteration order affects floating point summation which causes minor precision differences, that fail exact equality checks. NonDex output attached. 
[nondex_mimtest.txt](https://github.com/user-attachments/files/23831753/nondex_mimtest.txt)

### DIFF
--- a/Classification/FeatureSelection/src/test/java/org/tribuo/classification/fs/MIMTest.java
+++ b/Classification/FeatureSelection/src/test/java/org/tribuo/classification/fs/MIMTest.java
@@ -26,6 +26,7 @@ import org.tribuo.classification.Label;
 import org.tribuo.classification.LabelFactory;
 import org.tribuo.impl.ArrayExample;
 import org.tribuo.provenance.SimpleDataSourceProvenance;
+import org.tribuo.util.Util;
 import org.tribuo.util.infotheory.InformationTheory;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


Classification/FeatureSelection/src/test/java/org/tribuo/classification/fs/MIMTest.java
```diff
@@ -144,7 +144,10 @@ public void mimTest() {
         assertEquals(5,scores.size());
 
         assertEquals(Arrays.asList("E","B","C","D","A"),sfs.featureNames());
-        assertEquals(Arrays.asList(1.0, 0.1887218755408671, 0.1887218755408671, 0.0487949406953985, 0.0),sfs.featureScores());
+        List<Double> expectedScores = Arrays.asList(1.0, 0.1887218755408671, 0.1887218755408671, 0.0487949406953985, 0.0);
```
Craigacp:
```suggestion
        double[] expected = new double[]{1.0, 0.1887218755408671, 0.1887218755408671, 0.0487949406953985, 0.0};
        assertArrayEquals(expected, Util.toPrimitiveDouble(sfs.featureScores()), 1e-12);
```

This is a little shorter and will give a slightly better error message when the test fails.

Craigacp:
It looks like if we fix `DenseFSMatrix` to use `LinkedHashMap` then the mi values here need updating but it doesn't need the epsilon. Might still be worth adding to improve robustness anyway, the implementation of transcendental functions differs between different JVM implementations and architectures so it might be flaky for other reasons.

PR #423, opened 2025-11-29
Fix mRMR non-determinism with lexicographic tie-breaking

lbh930:
### Description
Modify mRMR to use lexicographic tie breaking when feature scores are within a floating point tolerance, so feature selection is deterministic when multiple features have identical or nearly identical scores.

### Motivation
NonDex showed 'mRMRTest' is flaky because the order of selected features depended on the iteration order of the underlying map when scores were tied. With a consistent tie-breaking rule (lexicographically by feature name in this PR), the algorithm should produce stable results regardless of ordering issues. A more conservative way to fix this test flakiness is to change mRMRTest only to ignore ordering. NonDex output attached. 

[nondex_mRMRTest.txt](https://github.com/user-attachments/files/23834967/nondex_mRMRTest.txt)


lbh930:
Hi, I agree that DenseFSMatrix is the root cause. I verified with NonDex that by fixing DenseFSMatrix with LinkedHashMap in this PR #422 also fixes mRMRTest.

PR #425, opened 2025-12-01
Fix flaky ReproUtilTest by adding fallback lookup for trainers

lbh930:
### Description
Modified ReproUtil.recoverTrainer to include a fallback when looking up trainer components. If lookup by name fails due to mismatched component names, it now attempts to look up the component by its class type. It sorts the candidate components by their numeric suffix to ensure deterministic mapping between the provenance and configuration list.

### Motivation
NonDex detected test flakiness in:
- `org.tribuo.reproducibility.ReproUtilTest.reproduceTransformTrainer`
-  `org.tribuo.reproducibility.ReproUtilTest#testBaggingTrainer `
- `org.tribuo.reproducibility.ReproUtilTest.testBaggingTrainerAllInvocationsChange`

One of the cause is possibly the non-deterministic component naming in OLCUT. ProvenanceUtil generates component names like logisticregressiontrainer-1 vs logisticregressiontrainer-2 based on iteration order, which relies on HashMap/HashSet and has no guarantee in determinism. The PR allow ReproUtil to recover the trainer based on its type when the name lookup fails and enforces a strict sorting order to guarantee the correct component is selected. This should improve robustness of those tests. An example of NonDex output attached.

[NonDex_ReproUtilTests.txt](https://github.com/user-attachments/files/23867352/NonDex_ReproUtilTests.txt)

### DIFF
--- a/Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
+++ b/Reproducibility/src/main/java/org/tribuo/reproducibility/ReproUtil.java
@@ -177,6 +177,7 @@ public final class ReproUtil<T extends Output<T>> {
         // construct the correct component name using ProvenanceUtil.computeName, and
         // link a provObject to its corresponding configuration in the cm.
         ProvenanceUtil.ProvenanceOrdering ordering = ProvenanceUtil.orderProvenances(this.modelProvenance);
+        Set<String> usedComponentNames = new java.util.HashSet<>();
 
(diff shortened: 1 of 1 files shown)


Craigacp:
I think I'd prefer to try and fix this in OLCUT (which is also maintained by our group), but that's lower down on the list of priorities. I'll look into it in the new year.

PR #437, opened 2026-05-07
Adding more serialization validation and tightening a few checks

Craigacp:
### Description
Adds a deserialization filter to the LibLinear models, modifies how OCIModel deserializes so that users need to supply the endpoint URL, and tightening a few shape checks in DenseMatrix.

### Motivation
Improve serialization behaviour.


JackSullivan (approved):
Approved

